PR #10, opened 2020-11-30
feat: react-navigation support

andreialecu:
This PR adds a `createMaterialCollapsibleTopTabNavigator` function which can be used with `react-navigation`.

Usage:

```tsx
import {createMaterialCollapsibleTopTabNavigator} from 'react-native-collapsible-tab-view';

const Tab = createMaterialCollapsibleTopTabNavigator();

export function YourTabScreen() {
  return (
    <Tab.Navigator
      collapsibleOptions={{
        tabBarHeight: 30,
        // disableSnap: true,
        headerHeight: 250,
        renderHeader: () => <CoverImage {...params} />,
      }}
   ...
```

Additionally, I made some hopefully minor project infrastructure changes:
* It seems that the root project was set up to use npm and had a `package-lock.json` but the scripts in `package.json` were using yarn. So I changed it to yarn.
* Added a `prepack` script so that it can be installed from git and properly compiled. (`prepack` is ran when installing a git dependency)
* In order for `prepack` not to fail because `bob build` also type checks the `example` folder, I excluded `example` from being compiled by the root and instead created its own `tsconfig.json`

If you have any questions or concerns, let me know. See individual commits for details.

It can be tested via:

`yarn add react-native-collapsible-tab-view@andreialecu/react-native-collapsible-tab-view#navigation`

(documentation still needs to be updated, and the example as well)


andreialecu:
PS: Allow edits from maintainers is on, feel free to adjust anything as necessary.

PedroBern:
Don't mind the expo preview action failing, it won't work because PR doesn't have access to my secrets env variables. Can you add an example to the example folder?

PedroBern:
Sorry for taking a long time to review this, I'm on a busy week, but will review it later

andreialecu:
@PedroBern I added an example. If you wouldn't mind updating the docs for it it would be great, you'd be doing a better job at it than me.

andreialecu:
It seems that #12 still had a minor issue if you initiated a momentum scroll gesture while the header was in the process of snapping back. The loop would still start and make the snapping janky. You can reproduce it by scrolling quickly while it's snapping back on `main`.

I have added a commit that reworks the debounce into a throttle, and removed the dependency that #12 added. It seems to work better overall, and is quicker to snap back.

I added it as part of this PR since it was easier to do, otherwise I need to keep merging `main` into this branch, hopefully that's ok.

PedroBern:
> @PedroBern I added an example. If you wouldn't mind updating the docs for it it would be great, you'd be doing a better job at it than me.

It's perfect, I will release after updating the docs

PedroBern:
@andreialecu I haven't tried to use the project without `react-navigation` after this PR, until now with #27. Looks like it's not an optional dependency. I tried, but don't know exactly how to conditionally export the navigator. Perhaps do you know? If you don't know, I will have to set it as a peer dependency without the `optional` meta field.

andreialecu:
Can you clarify what the error or behavior you are seeing is? 

The dependencies are added as optional via the `peerDependenciesMeta` field, which is supported by all package managers, so it should work properly as-is and indeed be optional.

Issue #9, opened 2020-11-30
Using `snapToOffsets` instead of the custom snap code


andreialecu:
I noticed that there's some sort of loop happening when doing very small momentum scrolls within the header. 

I'm not sure if this reproduces outside of my project, and I also use `react-navigation` via a WIP PR I'm working on for #7

What seems to be happening is that `onMomentumScrollEnd` triggers via the user gesture, then `syncScrollOffset` calls `scrollScene` which then also triggers `onMomentumScrollEnd` and then there's a battle between the two triggering a loop.

This seems to have also been reported here: https://stackoverflow.com/questions/61715531/abrupt-scrolltooffset-calls-in-onmomentumscrollend-react-native

However, it seems to be possible to use https://reactnative.dev/docs/scrollview#snaptooffsets with values of `[0, headerHeight]` to achieve the same behavior as the custom scroll snapping. Is there any reason not to use `snapToOffsets`?

A current workaround is to set `disableSnap` to true and add `snapToOffsets` to my own list views

andreialecu:
I guess one reason for using the custom snap code is the ability to override the `snapThreshold`. 

The native `snapToOffsets` prop has improved behavior and stops momentum scrolling at the snap points, so the header doesn't peek through before snapping back, which may be preferable.

andreialecu:
May need to verify that it doesn't break cases when the tab content slightly overshoots the window size by less than the snap offset, so it doesn't get stuck offscreen with the list snapping back and not allowing access to it.

PedroBern:
The `snapToOffsets` and `snapToInterval` indeed are a better solution for the `ScrollView`, but won't work for the `FlatList`. The current implementation is a simple way to handle both cases at the same time/same way.

PedroBern:
I did a quick test on the examples folder, adding `snapToOffsets={[0, HEADER_HEIGHT]}` to all scrollable components and running on the `Collapsible Tab View no snap` demo, wich has `disableSnap={true}`, and it looks like working fine for scroll view, but does not work for flatlist.

andreialecu:
Weird. It works with FlatList in the app I'm working on with no issues. I didn't test the examples folder though, something must be different.

andreialecu:
@PedroBern make sure to also add `snapToEnd={false}` otherwise the snap points keep repeating for the rest of the list, resulting in weird scrolling behavior.

PedroBern:
[...]

Issue #8, opened 2020-11-29
Header gets stuck if a tab doesnt have enough content


sniib:
First of all, thanks for creating this awesome wrapper!

As you can see, when I scroll down the A tab and switch to B, the header gets stuck and there is empty space between the header and the content. I guess the expected behaviour would be that the header resets to max height when switching to tab B.

![Alt Text](https://media.giphy.com/media/zgj2qA3BrRVGpveESk/giphy.gif)

andreialecu:
I'm also running into this and I'm curious how to solve it.

Since the user can see part of other tabs while swiping, I'm not sure how to visually handle this.

A potential solution might be to fix the gap up top so the tab content starts aligned to the top. Then once the gesture is released, the header might pop in and the tab content would move down.

PedroBern:
The above commit fix it, but a customized `minHeight`, including the navigation bar and status bar height, would be a better solution but must be provided in-app, not in-package.

PR #11, opened 2020-12-01
perf: use snapToOffsets instead of custom snap code

PedroBern:
related to #9

BREAKING CHANGE: remove snapThreshold prop

### DIFF
--- a/.github/workflows/expo-preview.yml
+++ b/.github/workflows/expo-preview.yml
@@ -43,14 +43,14 @@ jobs:
 
       - name: Get expo link
         id: expo
-        run: echo "::set-output name=path::@pedrobern/react-native-collapsible-tab-view-demos?release-channel=pr-${{ github.event.number }}
+        run: echo "::set-output name=path::@pedrobern/react-native-collapsible-tab-view-demos?release-channel=pr-${{ github.event.number }}"
 
       - name: Comment on PR
         uses: actions/github-script@v2
         with:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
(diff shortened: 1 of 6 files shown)


andreialecu:
Hey @PedroBern, I took a look at this and it seems the reason it works with `minHeight: 0` is because `minHeight` is being set to `1012`, which is too big, by the change in https://github.com/PedroBern/react-native-collapsible-tab-view/commit/98f7b7e4f44e25f9c225d29782c8d55fabdbc83f#diff-9cbdb93af845f8d4c83a1783ce090969769880a9f8f4f9fecf504ae01f1055e3R56

The bug is introduced by that commit. By the way, instead of using `Dimensions` I think it is preferrable to use the `useWindowDimensions()` hook: https://reactnative.dev/docs/usewindowdimensions just in case the dimensions update somehow (maybe by rotating the screen)

andreialecu:
One thing that isn't great with `snapToOffsets` is that the initial momentum scrolling when swiping down seems to be stopped at the snap point, so it needs to be scrolled twice to get further down than the initial point. 

For some reason this doesn't seem to happen on the iOS simulator on my end, but just on a real device.

andreialecu:
I was able to fix this with the current snapping code in #12 - which I prefer, because of the behavior in the comment above.

PedroBern:
> Hey @PedroBern, I took a look at this and it seems the reason it works with minHeight: 0 is because minHeight is being set to 1012, which is too big, by the change in 98f7b7e#diff-9cbdb93af845f8d4c83a1783ce090969769880a9f8f4f9fecf504ae01f1055e3R56

> The bug is introduced by that commit, and seems to also happen without the changes in this PR (visible on main).

> PS: instead of using Dimensions I think it is preferrable to use the useWindowDimensions() hook: https://reactnative.dev/docs/usewindowdimensions just in case the dimensions update somehow (maybe by rotating the screen)

@andreialecu  Instead of fixing it (#8), I'm adding an example of how to solve it #13

PedroBern:
closing in favor of #12

PR #12, opened 2020-12-02
fix: snapping when momentum scrolling

andreialecu:
What was happening was that `onScrollEndDrag` was calling `syncScrollOffset` when the user lifts the finger off the screen, but `onMomentumScrollBegin/onMomentumScrollEnd` is called immediately after and calls the same function. Since the function initiates its own scrolling, there's a loop of events going on.

See https://reactnative.dev/docs/scrollview#onscrollenddrag
> Called when the user stops dragging the scroll view and it either stops **or begins to glide.**

This PR adds a debounce effect so that it is only called once, and it should solve any performance issues without requiring to use `snapToOffsets`.

Will add a screen recording below.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -31,7 +31,8 @@
     "babel-plugin-module-resolver": "^4.0.0",
     "babel-preset-expo": "^8.2.3",
     "expo-cli": "^3.23.1",
-    "typescript": "~3.9.7"
+    "typescript": "~3.9.7",
+    "use-debounce": "^5.1.0"
   },
   "resolutions": {
     "@babel/runtime": "^7.9.0"
--- a/example/yarn.lock
+++ b/example/yarn.lock
@@ -12981,6 +12981,11 @@ url@^0.11.0:
     punycode "1.3.2"
     querystring "0.2.0"
 
+use-debounce@^5.1.0:
+  version "5.1.0"
+  resolved "https://registry.yarnpkg.com/use-debounce/-/use-debounce-5.1.0.tgz#3df34bd7449da7dd2dbb10a6513efe701fc79d02"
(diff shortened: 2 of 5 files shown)


andreialecu:
Broken:

broken

Fixed:

fixed

PedroBern:
@andreialecu now its much better user experience, it feels really smooth!

andreialecu:
There is still a glitch here when doing micro-scrolls when the header is within partially visible. It's much better than before but the event loop issue can still be seen in certain edge cases. 🤔

PedroBern:
> There is still a glitch here when doing micro-scrolls when the header is partially visible. It's much better than before but the event loop issue can still be seen in certain edge cases. 🤔

@andreialecu I guess it wouldn't happen with `snapToOffsets`, since the sync scroll function would run only for unfocused routes. What do you think?

andreialecu:
That is correct, however I think the glitch can be fixed by adjusting how often `scrollScene` is called. Just need to find a better way to handle the throttling.

The reason it glitches is because `scrollScene` starts being called repeatedly, but we have control over when we call it. Just need to find a better way to control it. 🤔 

The remaining glitch, right now, only happens when the momentum scrolling is "settling" and the user scrolls by a very small amount.

The problem with `snapToOffsets` is related to how it always seems to snap/stop to the first point on the real iOS device, requiring two scrolls to get past it.

But the good news is that it is easy to switch to `snapToOffsets`. A user could set `disableSnap: true` and set his own `snapToOffsets: [0, HEADER_HEIGHT]` and `snapToEnd: false`.

PedroBern:
This glitch is so small, don't bothers me, at this point the package is on higher quality level compared to before your PRs, for me it's fine. And yes, super easy to switch. Anyway, if you find a way to improve, please send a new PR 😃

PR #15, opened 2020-12-03
feat: add `snapTimeout`

andreialecu:
This adds a configurable `snapTimeout` prop which allows setting the timeout after which snapping will occur.

I have also refactored the code a bit to fix the remaining glitches mentioned in https://github.com/PedroBern/react-native-collapsible-tab-view/pull/12#issuecomment-737434326


PedroBern:
I did a quick test here, and I feel now it's making an abrupt jump to snap, see below. The current implementation is so smooth and natural.

ezgif com-gif-maker

andreialecu:
That is weird, is that on iOS or Android? It should behave the same, but I didn't test Android.

PedroBern:
It's Android, and I didn't test iOS!

andreialecu:
I pushed a commit that fixes the android behavior. Also refactored things a little.

PedroBern:
The snap effect got really better, but it introduced 1 new bug and make another old bug more visible:

#### 1. Old bug becoming more visible
- Start scrolling
- Stop scrolling
- Before the snap effect occur, start scrolling again
- **Expected**: cancel the snap
- **Behavior**: the snap occurs, even with the finger on the screen. But it's super rare since the snap takes effect almost immediately after the "first stop scrolling"

But I'm more concerned about the new bug:

#### 2. New bug

- Click on a tab
- Start scrolling immediately
- **Expected**: scroll shouldn't work (current) / or the header must follow the screen (desired)
- **Behavior**: scroll works and the header doesn't follow the scroll

See the demo of this new bug:

ezgif com-gif-maker (1)

andreialecu:
Hey, so I looked into this and:

1. On Android `onScrollBeginDrag` has a slight delay of a few milliseconds before it triggers and we can't cancel the effect in time because of it so this is more visible there. You can see by adding a `console.log` to the event - it triggers late. On iOS that event triggers instantly. It's possible to work around it using a longer `snapTimeout`. **However, I see same behavior on the `main` branch.**

2. I think the latest commit fixes this.

andreialecu:
Noticed an additional edge case, working on it.

andreialecu:
Hey @PedroBern, check out the latest commit. It should hopefully fix any issues with switch tabs after scrolling.

It seems that the reason that on `main` swiping is locked while snapping is just a side effect of programatically scrolling the scroll views. That's a behavior that isn't guaranteed, and mostly seems like a react native bug. I was not able to determine any other reason why the swiping gets disabled besides this.

But the new commit should have improved behavior, with the new tab starting right under the header and snapping itself to the right position.

PedroBern:
> ...However, I see same behavior on the main branch.

It's true.

> But the new commit should have improved behavior...

Unfortunately, I'm still experiencing the same bug :(

Also noticed other unexpected behavior, when doing the following:

1. Scrolling up to half header
2. Remove finger
3. Before the timeout scroll all the alway up and a little more
4. The tab is scrolled down, what's weird, because we are still scrolling up.

To illustrate, I recorded this bug in slow motion (second gif), showing the touches, notice how I'm still with the finger on the screen when it's snapping.

regular


slow

PedroBern:
> It seems that the reason that on main swiping is locked while snapping is just a side effect of programatically scrolling the scroll views. That's a behavior that isn't guaranteed, and mostly seems like a react native bug. I was not able to determine any other reason why the swiping gets disabled besides this

I don't see the swiping beeing locked, actually I can swipe while snapping on main.

andreialecu:
The problems you mentioned are due to the problem with events firing late on Android. Mainly the ones to detect if a user started scrolling. I think they're both the same problem. 

Not sure if it is more visible on the simulator or it happens on real devices as well. I don't have an android device to test.

Unfortunately without reliable event firing this is pretty hard to fix. I have an idea though, I'll play with it some more tomorrow.

andreialecu:
Hey @PedroBern, after a bunch of trial and error with various approaches I think the latest commit handles things much better on Android as well. I have tested it on some real devices and it seems to work well. Let me know what you think.

PedroBern:
Hey @andreialecu, first of all, thanks for all the greatest contributions!

I have 2 android devices, one old and one new. The old have a [geekbench5](https://www.geekbench.com/) score of 102 single-core and 377 multi-core, pretty bad! And I'm still getting the "tab-switching-gap" bug 😔 , but on the new one it's perfect, you nailed it 👏 

So I have this dilemma, currently, I don't have the bug on the main gap when running on the old device. So, should we sacrifice the user experience for older devices? I see and agree with the improvements, but what's the issue with the current implementation? I don't have an iOS device, so please walk me through if it's iOS specific. I remember you mentioning a small glitch.

andreialecu:
There's currently a bug on `main` that can happen when the user starts scrolling for a tiny bit while the header is still visible. It results in an infinite loop which can also drain battery.

I can't attach it here since it would be too big for a gif, but you can take a look at a recording I made here:
https://drive.google.com/file/d/1zNRILj_afpDecgYZjW7DvYDUXqx6r_Ed/view?usp=sharing

Notice the `console.log` I added and the Terminal. Also the bug is also visible since the snap effect is very slow. What happens is that it scrolls for a few pixels at a time and re-enters the loop.

PedroBern:
@andreialecu wow, that was awful! Forget old devices, thanks for all the patience! I will merge it now. Are there any changes you want to make before merging?

andreialecu:
I think 250 is pretty smooth and there's little difference perception-wise vs 100, do you feel it's slow?

I used 250 because Android seems to have a small delay before it triggers the various events, so 250ms gives it more time so they can get handled in certain cases. I didn't see an issue on iOS at either 100 or 250, but either way the behaviour feels pretty smooth to me, and similar to the delay other apps seem to have before snapping.

andreialecu:
That said, I didn't find any other issues at the moment that I'm working on. So the PR is ready to merge from me 🚀. If I find anything else, it can come along in a separate PR.

andreialecu:
Oh, one comment that needs fixing. Perhaps you can apply the change from the UI directly, I don't have the repo open currently to fix it.

PedroBern:
@andreialecu 250 is perfect, I will release it now! 🚀

PR #30, opened 2020-12-28
feat: scroll on header

### LINKED ISSUE #28: Respond to scroll gestures on collapsible header
### Feature request
In some cases, where the collapsible header covers a significant amount of the vertical screen space, a user might prefer or even be forced to start their scroll gesture on the collapsible header. There are a few cases I can think of where this may happen:

- **When the header is used to display larger amounts of content**. Say you are developing a delivery app and want to display some pictures and information about a restaurant above a tab list of all their dishes. The header will initially cover most of the screen (see Illustration).
- **When using a small screen device**. While phones with small screens admittedly seem to be a dying breed a bunch of them are still out there. Due to different scaling and aspect ratios most headers would likely cover >50% of the screen.
- **When using split screen mode on Android devices**. Even regular sized collapsible headers like in the example app will cover almost all of the available window height (see Screenshot).

Currently, when starting a scroll gesture on the header nothing happens at all. It would be nice if there was a (opt-in) way for these gestures to be passed or translated to the `ScrollView` / `FlatList` inside the focused tab. This behavior would also be consistent with some native implementations collapsible headers (e.g the "Galaxy Store" App).

If anyone has ideas on how this could be implemented I would love to help and open a PR.

### Screenshots
| Large header  | Android Split Screen |
| ------------- | ------------- |
| Header with large amounts of content  | split_screen|

PedroBern:
closes #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@
 - [Features](#features)
 - [Installation](#installation)
 - [Quick Start](#quick-start)
+- [Scroll on header](#scroll-on-header)
 - [API reference](#api-reference)
   - [CollapsibleTabView](#collapsibletabview)
   - [useCollapsibleScene](#usecollapsiblescene)
@@ -120,8 +121,10 @@ const SecondScene = () => <SomeRoute routeKey="second" color="black" />;
 
 const HEADER_HEIGHT = 250;
 
+// set pointerEvents="none" to allow scroll on header
+// see the docs for more information
 const renderHeader = () => (
-  <View style={styles.header}>
+  <View pointerEvents="none" style={styles.header}>
     <Text style={styles.headerText}>COLLAPSIBLE</Text>
   </View>
 );
@@ -193,6 +196,14 @@ const styles = StyleSheet.create({
 
 export default WithReactNavigation;
 ```
+
+## Scroll on header
+
+If you want to allow scrolling from the header:
+
+- If `renderHeader` **doesn't** contain touchables set `pointerEvents='none'`
+- If `renderHeader` **does** contain touchables set `pointerEvents='box-none'` for them to work.
+  _Note: With this setting any child component that should **not** respond to touches (e.g. `<Image />`) needs to have `pointerEvents` set to `'none'`. Otherwise it can become the target of a touch gesture on iOS devices and thereby preventing scrolling._
 
 ## API reference
 
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -8,7 +8,6 @@ import {
   Text,
   TouchableOpacity,
   View,
-  YellowBox,
   I18nManager,
 } from 'react-native';
 import { registerRootComponent } from 'expo';
@@ -18,6 +17,7 @@ import Constants from 'expo-constants';
 import { Ionicons } from '@expo/vector-icons';
 import AsyncStorage from '@react-native-community/async-storage';
 import CollapsibleTabViewExample from './CollapsibleTabViewExample';
+import CollapsibleTabViewScrollOnHeaderExample from './CollapsibleTabViewScrollOnHeaderExample';
 import CollapsibleTabViewSmallContentExample from './CollapsibleTabViewSmallContentExample';
 import CollapsibleTabViewNoSnapExample from './CollapsibleTabViewNoSnapExample';
 import CollapsibleTabViewResizeExample from './CollapsibleTabViewResizeExample';
@@ -35,8 +35,6 @@ type State = {
 
 I18nManager.forceRTL(false);
 
-YellowBox.ignoreWarnings(['bind():']);
-
 const PERSISTENCE_KEY = 'index_persistence';
 
 const EXAMPLE_COMPONENTS: ExampleComponentType[] = [
@@ -48,6 +46,7 @@ const EXAMPLE_COMPONENTS: ExampleComponentType[] = [
   CollapsibleTabViewNoUpfrontHeightExample,
   MaterialTopTabsCollapsibleTabViewDemoExample,
   CollapsibleTabViewCenteredEmptyListExample,
+  CollapsibleTabViewScrollOnHeaderExample,
 ];
 
 const KeepAwake = () => {
--- a/example/src/CollapsibleTabViewScrollOnHeaderExample.tsx
+++ b/example/src/CollapsibleTabViewScrollOnHeaderExample.tsx
@@ -1,0 +1,13 @@
+import React from 'react';
+import CollapsibleTabView from './CollapsibleTabViewExample';
+import { ExampleComponentType } from './types';
+
+const CollapsibleTabViewExample: ExampleComponentType = () => {
+  return <CollapsibleTabView enableScrollOnHeader />;
+};
+
+CollapsibleTabViewExample.title = 'Scroll on header';
+CollapsibleTabViewExample.backgroundColor = '#2196f3';
+CollapsibleTabViewExample.appbarElevation = 0;
+
+export default CollapsibleTabViewExample;
--- a/example/src/CollapsibleTabViewScrollOnHeaderExample.tsx
+++ b/example/src/CollapsibleTabViewScrollOnHeaderExample.tsx
@@ -1,9 +1,81 @@
-import React from 'react';
-import CollapsibleTabView from './CollapsibleTabViewExample';
+import * as React from 'react';
+import { StyleSheet, View, Text } from 'react-native';
+import { SceneMap } from 'react-native-tab-view';
+
+import {
+  CollapsibleTabView,
+  useCollapsibleScene,
+  CollapsibleTabViewProps,
+} from 'react-native-collapsible-tab-view';
+
+import { AnimatedAlbums } from './Shared/Albums';
+import { AnimatedArticle } from './Shared/Article';
+import { AnimatedContacts } from './Shared/Contacts';
 import { ExampleComponentType } from './types';
 
-const CollapsibleTabViewExample: ExampleComponentType = () => {
-  return <CollapsibleTabView enableScrollOnHeader />;
+type Route = {
+  key: string;
+  title: string;
+};
+
+export const ContactsScene = () => {
+  const scenePropsAndRef = useCollapsibleScene('contacts');
+  return <AnimatedContacts {...scenePropsAndRef} />;
+};
+
+export const ArticleScene = () => {
+  const scenePropsAndRef = useCollapsibleScene('article');
+  return <AnimatedArticle {...scenePropsAndRef} />;
+};
+
+export const AlbumsScene = () => {
+  const scenePropsAndRef = useCollapsibleScene('albums');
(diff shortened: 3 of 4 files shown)


yfunk:
@PedroBern Thanks for opening a PR so fast, I will have a look at the changes later to see if I can find a solution for the bug you mentioned in the issue.

andreialecu:
@PedroBern I took a look at this and the following fixes the glitch from what I can tell:

```patch
diff --git a/src/CollapsibleTabView.tsx b/src/CollapsibleTabView.tsx
index 86c7116..3927f62 100644
--- a/src/CollapsibleTabView.tsx
+++ b/src/CollapsibleTabView.tsx
@@ -129,6 +129,7 @@ const CollapsibleTabView = <T extends Route>({
   const lastInteractionTime = React.useRef(0);
 
   const [canSnap, setCanSnap] = React.useState(false);
+  const panOffset = React.useRef(0);
 
   const activateSnapDebounced = useDebouncedCallback(
     () => {
@@ -341,6 +342,12 @@ const CollapsibleTabView = <T extends Route>({
 
   const panResponder = React.useRef(
     PanResponder.create({
+      onPanResponderGrant: () => {
+        const curRouteKey = routes[index][
+          routeKeyProp as keyof Route
+        ] as string;
+        panOffset.current = listOffset.current[curRouteKey] || 0;
+      },
       onPanResponderRelease: () => {
         isGliding.current = false;
         lastInteractionTime.current = Date.now();
@@ -361,7 +368,7 @@ const CollapsibleTabView = <T extends Route>({
           if (isCurrentRoute) {
             scrollScene({
               ref: item.value,
-              offset: -gs.dy,
+              offset: panOffset.current + -gs.dy,
               animated: false,
             });
           }

```

However, could this be behind a flag, or at least allow it to be disabled? 

I haven't tested it in a real app, but enabling it by default may break some use cases depending on what people might do with their headers. 🤔

andreialecu:
Noticed that when switching tabs it stops working, but that seems like a separate bug.

yfunk:
> However, could this be behind a flag, or at least allow it to be disabled?

I would definitely put it behind a flag in the `CollapsibleTabView` props (maybe something like `enableScrollOnHeader`) and probably leave it disabled by default so as not to break existing implementations.

PedroBern:
@andreialecu thanks for the pointers! Solved partially, were really helpful.

> Noticed that when switching tabs it stops working, but that seems like a separate bug.

Fixed.

> I would definitely put it behind a flag in the CollapsibleTabView props (maybe something like enableScrollOnHeader) and probably leave it disabled by default so as not to break existing implementations.

Done 👍 

@yfunk please test the expo preview and let me know your thought. Can you test it on iOS?

Now it's working without bugs, but once we release the finger from the screen it stops, would be better if we could propagate the momentum of the gesture to the scroll. @andreialecu @yfunk do you know how to do it with panResponder ?

yfunk:
>Please test the expo preview and let me know your thought. Can you test it on iOS? There is a new example `9. Scroll on header`. It scrolls only in the header, not in the tabs.

I tested the expo preview on Android and iOS and the previously mentioned bugs all seem to be fixed. Overall it seems pretty solid, but I noticed a few small things that may need to be fixed:
- When starting a scroll gesture on the `tabBar` part of the collapsible header (no matter if it is collapsed or expanded) the list does not scroll. This was working before the most recent commit (2eb1aa9).
- Scrolling down on the header doesn't trigger the RefreshControl (although this may not even be possible to achieve).
- When scrolling down on a fully expanded header, not letting go and then scrolling back up again the list only starts moving again when the point where the gesture started was passed. RN components like `ScrollView` start scrolling again right when changing the direction

> Now it's working without bugs, but once we release the finger from the screen it stops, would be better if we could propagate the momentum of the gesture to the scroll.

That would be a nice feature, but unfortunately I don't know how it could be implemented. But I will definitely look into it and the other small issues tomorrow to see if I can figure something out.

andreialecu:
One other (simpler) way to implement this which retains the the refresh control and momentum scrolling:

```
diff --git a/example/src/CollapsibleTabViewExample.tsx b/example/src/CollapsibleTabViewExample.tsx
index a38f3f2..ceeeee3 100644
--- a/example/src/CollapsibleTabViewExample.tsx
+++ b/example/src/CollapsibleTabViewExample.tsx
@@ -36,7 +36,7 @@ export const AlbumsScene = () => {
 export const HEADER_HEIGHT = 250;
 
 const renderHeader = () => (
-  <View style={styles.header}>
+  <View pointerEvents="none" style={styles.header}>
     <Text style={styles.headerText}>COLLAPSIBLE</Text>
   </View>
 );
diff --git a/src/CollapsibleTabView.tsx b/src/CollapsibleTabView.tsx
index 46721cf..86d8764 100644
--- a/src/CollapsibleTabView.tsx
+++ b/src/CollapsibleTabView.tsx
@@ -354,6 +354,7 @@ const CollapsibleTabView = <T extends Route>({
   ): React.ReactNode => {
     return (
       <Animated.View
+        pointerEvents="box-none"
         style={[
           styles.headerContainer,
           { transform: [{ translateY }] },

```

I haven't tested it too much so an expo preview would be good.

andreialecu:
The following retains `Press to resize` button functionality, but scrolling can't be initiated while touching it:

```patch
diff --git a/example/src/CollapsibleTabViewResizeExample.tsx b/example/src/CollapsibleTabViewResizeExample.tsx
index bb45e61..2b4172a 100644
--- a/example/src/CollapsibleTabViewResizeExample.tsx
+++ b/example/src/CollapsibleTabViewResizeExample.tsx
@@ -25,7 +25,7 @@ const Header: React.FC<{ height: number; onPress: () => void }> = ({
   onPress,
 }) => {
   return (
-    <View style={[styles.header, { height }]}>
+    <View pointerEvents="box-none" style={[styles.header, { height }]}>
       <TouchableOpacity onPress={onPress}>
         <Text style={styles.headerText}>Press to resize</Text>
       </TouchableOpacity>
```

yfunk:
@andreialecu The `pointerEvents` approach seems to work very well on Android, but unfortunately not at all on iOS (as far as I know you can't perform touch gestures on a view that is under another solid or even transparent view at all on this platform).

> The following retains `Press to resize` button functionality, but scrolling can't be initiated while touching it (which is probably fine)

I think it would be important for this to be possible like it was in 02abd1b (gesture initially registers as touch on touchable and is cancelled when a scroll gesture starts). If you have touchables with `width: 100%`  in your header (going back to my delivery app example this could be a preview of a description text that opens a new screen with the whole description when pressed) a user is likely to start their touch gesture somewhere on there and will be confused as to why nothing happens.

andreialecu:
> @andreialecu The pointerEvents approach seems to work very well on Android, but unfortunately not at all on iOS (as far as I know you can't perform touch gestures on a view that is under another solid or even transparent view at all on this platform).

@yfunk this is weird, I came up with this solution while testing on iOS. I didn't test Android at all. Seems to work fine on my end. 🤔

yfunk:
@andreialecu Sorry, my fault. I made the changes in my working copy to quickly test them and must have missed something. I checked out `main` again, applied your diffs and now its working on both platforms. 

Although I noticed, that on iOS when `renderHeader` is set to set to `pointerEvents="box-none"` **any** child like a `Text` or `Image` component (even if not explicitly declared touchable) will become the target of a touch gesture and prevent scrolling. You can circumvent this by setting `pointerEvents="none"` on each one individually, but this is getting in a territory where this would be more of a hack than a library feature.

andreialecu:
It may be possible to use the [gesture responder system](https://reactnative.dev/docs/gesture-responder-system) somehow to prevent the header from taking control of the pan responders. 

I wasn't able to get it to work via some quick experimentation, presumably because the header is absolutely positioned over the scroll view and it doesn't seem to bubble the events to the responder below it (the scroll view). This likely happens because they're not linked via a child/parent relationship.

Perhaps there's a way to get it to work, ideally the header just needs to disable its gesture handlers and instead relay them to the scroll view underneath.

Maybe this could help somehow:
https://github.com/simonhoss/react-native-touch-through-view


Note however that trying to achieve this functionality via a `PanResponder` like in [`2eb1aa9` (#30)](https://github.com/PedroBern/react-native-collapsible-tab-view/pull/30/commits/2eb1aa92fc74df6cd9a1539476ec9370cc982aa9) has some drawbacks:
- momentum scrolling will not work, and even if it can be simulated to work, it won't really feel native
- iOS elastic scrolling and bouncing at the start/end will also not work
- the refresh control won't work

PedroBern:
> One other (simpler) way to implement this which retains the refresh control and momentum scrolling

you solved in 1 line what I took 1 hour to accomplish with panResponder.

> Maybe this could help somehow:
https://github.com/simonhoss/react-native-touch-through-view

This won't work with Expo 😞 

I think now it's working good enough to release. Can you guys test the expo preview?

> Although I noticed, that on iOS when renderHeader is set to set to pointerEvents="box-none" any child like a Text or Image component (even if not explicitly declared touchable) will become the target of a touch gesture and prevent scrolling. You can circumvent this by setting pointerEvents="none" on each one individually, but this is getting in a territory where this would be more of a hack than a library feature.

Yes, it could become really boring to set `pointerEvents="none"` for each child, but I think in most cases, the header will be an image or text, not a complex component (my use case for this lib is just text). Usually, the action buttons will be above, in the screen header (if for example using react-navigation), so I guess it's fine.

yfunk:
> I think now it's working good enough to release. Can you guys test the expo preview?

Aside from what I mentioned previously it seems to work pretty well. Maybe you should add some documentation about correctly setting `pointerEvents` in the `renderHeader` component:

- If `renderHeader` **doesn't** contain touchables set `pointerEvents='none'`
- If `renderHeader` **does** contain touchables set `pointerEvents='box-none'` for them to work. 
 _Note: With this setting any child component that should **not** respond to touches (e.g. `<Image />`) needs to have `pointerEvents` set to `'none'`. Otherwise it can become the target of a touch gesture on iOS devices and thereby preventing scrolling._

Unfortunately this won't really work for my use case as I have a bunch of touchables in my header. I'll see if I can find a solution that works for me, but you can definitely still release this change for other people to use.

andreialecu:
FWIW I just noticed that the Reddit app on iOS has a similar collapsible header and tab view. It also does not allow initiating a scroll gesture from any of the touchable controls in the header, but it works from purely presentational components.

PedroBern:
> Unfortunately this won't really work for my use case as I have a bunch of touchables in my header. I'll see if I can find a solution that works for me, but you can definitely still release this change for other people to use.

😞  sorry for this, please if you find a solution, share it here, or if have time, open a new PR 

> FWIW I just noticed that the Reddit app on iOS has a similar collapsible header and tab view. It also does not allow initiating a scroll gesture from any of the touchable controls in the header, but it works when initiated from purely presentational components.

On Android, the Reddit app has a tab that collapses, instead of the header. It allows initiating the scroll from any part, including the touchable components.

PR #37, opened 2021-01-06
feat: creating an example to show the problem when the FlatList needs to be nested

alexpchin:


alexpchin:
Closing as this is v2 related.

PR #40, opened 2021-01-13
feat: no header

alexpchin:
Please check whether it works.

### DIFF
--- a/src/CollapsibleTabView.tsx
+++ b/src/CollapsibleTabView.tsx
@@ -325,8 +325,8 @@ const CollapsibleTabView = <T extends Route>({
         setHeaderHeight(Math.max(value, 0));
         setTranslateY(
           scrollY.current.interpolate({
-            inputRange: [0, Math.max(value, 0)],
-            outputRange: [0, -value],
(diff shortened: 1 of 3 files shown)


src/CollapsibleTabView.tsx
```diff
@@ -322,11 +322,11 @@ const CollapsibleTabView = <T extends Route>({
       const value = event.nativeEvent.layout.height - tabBarHeight;
       if (Math.round(value * 10) / 10 !== Math.round(headerHeight * 10) / 10) {
         onHeaderHeightChange?.();
-        setHeaderHeight(Math.max(value, 0));
```
PedroBern, line 325: We need to make sure that the value is positive, otherwise, the Animated API will throw an error. That's why we pick the maximum value, between `value` and 0.

alexpchin: Doesn't this do that, if not - it defaults back to tabBarHeight?

alexpchin: Sorry @PedroBern you were right here. I think this does need to remain like this.

PR #41, opened 2021-01-14
v3 - rework

### LINKED ISSUE #2: Allow passing ref to control currently focused scene
Currently, there is no way to pass a `ref` to the scrollable components. Would be nice to support it.

### LINKED ISSUE #19: Gap between header and content if lazy === true
### Current behavior

`<CollapsibleTabView  lazy ...>` do not sync scroll position of unmounted routes, as expected, because the routes are unmounted! But, when the routes are mounted, may start with a gap between the header and the content.

### Expected behaviour

When mounting the lazy routes, they should know the header position and start with the correct scroll position.

### Workaround

The current workaround is to handle the laziness by yourself. What actually can be a good thing, since the lazy prop of `react-native-tab-view` is pseudo-lazy. It loads all routes if you click on the last tab!

### LINKED ISSUE #20: Shared Pull To Refresh
### Feature request

You may want to implement a pull to refresh that is shared across all tabs. This might just be good to be included in the examples.

### Current behavior

Currently, the examples load a pullToRefresh from each tab's scene independently. I.e. if you do not implement a pull to refresh on each tab, then depending on what tab you are on you you'll get different behaviour.

TAB
- Scene1: FlatList with images
- Scene2: Static user information
- Scene3: ScrollView with other information

### Proposed solution

A shared pullToRefresh that implements a "refreshing" value that can be used trigger a `useEffect` in each scene.

### LINKED ISSUE #32: Padding showing when changing tabs.
Thank you for your great library!

When I have scrolled down on one tab so that the header is hidden and then change to another tab (via swipe or tap on the tab) that is scrolled all the way to the top, the paddingTop is showing.  It would usually be obscured by the header and tabs but the header is shrunk because of the interaction with the previous tab.

Is there known solution for this? Or a canonical approach? 
Thank you very much for your time.

### LINKED ISSUE #38: Allow a hook like `react-navigation` hook `useScrollToTop` to scroll all tabs to the top
A feature request to enable a ref on the `CollapsibleTabView` when clicking on the `react-navigation` bottom tab, to reset the scroll of all of the tabs to the top.

PedroBern:
@andreialecu @alexpchin 

rework with reanimated v2


alexpchin:
Hi @PedroBern Before integrating this`react-native-collapsible-tab-view`, I started doing something similar to this. Initial thoughts (not problems, just thoughts).

- Scrolling on the header will only work when you touch the Header, if the header included Touchables, it might cause a problem?
- Due to `nestedScrollEnabled`, the inner scroll will need to be connected to the outer scroll.
- I thought there was a warning about `VirtualizedLists should never be nested inside plain ScrollViews` but as the components are Animated, this might not apply

I will look in more detail later, I'm trying to implement the current version into an app (without any bugs) so that it can be pushed live. Once that is done, I can come back to this!

PedroBern:
> Scrolling on the header will only work when you touch the Header, if the header included Touchables, it might cause a problem?

The recent push can scroll anywhere. Don't about touchables yet.

> I thought there was a warning about VirtualizedLists should never be nested inside plain ScrollViews but as the components are Animated, this might not apply

It does not apply here because we have the following structure:

_ Vertical Scrollview
____ Horizontal Flatlist <- would have the warning if we don't have this intermediate component
________ Vertical Flatlist 

Check out the new expo preview, now there are 2 examples, in the second one we can see all tabs at once, which makes it easy to develop.

alexpchin:
_ Vertical Scrollview
____ Horizontal Flatlist <- would have the warning if we don't have this intermediate component
________ Vertical Flatlist

Isso é muito sorrateiro.

PedroBern:
@alexpchin removed the top level scrollview. Now it is very performant, (not reusable yet, just a demo).

I removed the sync scroll offsets. It's translating just the header now. Not sure if it's good or bad.

PedroBern:
I'm canceling this PR.

PedroBern:
sorry for taking your time

alexpchin:
@PedroBern No problem!

PedroBern:
@alexpchin @andreialecu v3 is ready, just need to update the README, now all animations/scrolls are done in the UI thread with reanimated.

Can someone please run the preview on iOS and give me feedback? I tested only on android.

New features:
- diffClamp
- diffClamp with animated snap (instead of scroll snap)

Removed
- react-native-tab-view
- react-navigation support (sorry @andreialecu)

andreialecu:
I can take a look in a couple of days. Would probably stay on v2 for the current project though as that has been working well so far.

Lacking react navigation support would be a blocker anyway.

Do you plan on adding it back later?

PedroBern:
That would be nice, thanks.

We definitely can add it back later.

Right now the v3 does not even come with a tab bar, the user must provide a custom one, but there is a nice example in the examples folder, a simplified version of the material one.

But the performance gains were fantastic, all computations are done in the UI thread now. Including the scroll sync and snapping. In v2 all these computations are done in the JS thread and that was causing a bug where the tabs do not sync. It was reported for at least 4 people and me, one has even sent a reproducible snack.

alexpchin:
Going to review this tomorrow, time permitting. Nice work @PedroBern

PedroBern:
Now we have real lazy support in v3.2

With react-native-tab-view it was an issue.

PR #42, opened 2021-01-14
Feat no header

alexpchin:
Hi @PedroBern There were still a few issues even after this fix after implementing the new version.

Firstly, you were correct with:

```ts
setHeaderHeight(Math.max(value, 0));
```

However, there was a little more to it.
When using the default `tabBarHeight` of 49, the TabBar was animating 1px down on scroll. Please see screenshot:

Screenshot 2021-01-14 at 11 12 30

I believe this is because the height of the tabBar was being measured by `getHeaderHeight` as 48px rather than 49px. I think this is because `react-native-tab-view` now uses a minimum TabBarItem height of 48px:

https://github.com/satya164/react-native-tab-view/blob/main/src/TabBarItem.tsx#L258

After updating the default `tabBarHeight` to `48`, this line had to be removed:

```ts
if (Math.round(value * 10) / 10 !== Math.round(headerHeight * 10) / 10) {
```

As it will be true (48 - 48 = 0) and if no header is provided, then `headerHeight` will also be 0. 

Then finally, I passed `tabBarHeight` out via `useCollapsibleScene` so that `tabBarHeight` can be used instead of a magic number.

Please take a look when you have a moment.


alexpchin:
@PedroBern I think this PR is ok? If you could check that would be 👍  as I am waiting to deploy with this fix

PedroBern:
@alexpchin I'm fetching it right now to see locally

PedroBern:
@alexpchin already published to npm 👍

alexpchin:
Nice one @PedroBern

PR #43, opened 2021-01-16
feat: is it possible to prevent re-renders?

alexpchin:
WIP: Just an idea. Can we prevent unnecessary re-renders?

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -14,6 +14,7 @@
     "@react-native-community/async-storage": "~1.12.0",
     "@react-navigation/material-top-tabs": "^5.3.10",
     "@react-navigation/native": "^5.8.10",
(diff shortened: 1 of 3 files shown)


PedroBern:
@alexpchin I reopened #41 and now I think it's going to really improve the performance. check out the new demo, it's super simple compared to the current implementation and much more bug-free. The downside is it does not come with a tab bar by default, but it's going to be super simple to provide a custom one.

This isn't my priority right now, but I'm going to try to update it in the next few days and make a reusable component.

I'm commenting here so maybe you don't waste too much time trying to improve the current implementation, preventing re-renders.

If you have something to add to the PR now or in the future, please let me know

alexpchin:
Closing as this is v2 focused.

PR #56, opened 2021-01-26
optimize demo images

sayem314:
the homepage takes forever to load so I have optimized all gif without visual loss

previous size for all gif was 55MB
new size is 26MB


PedroBern:
Hi @sayem314 thanks for the help 👍

PR #58, opened 2021-01-28
Feat react navigation

### LINKED ISSUE #64: Need option for changing Tab Bar Background Color
### Feature request

Need Option for changing Tab Bar Background Color

### Current behavior

Currently default white color hard coded in the stylesheet. Please add an option.

### LINKED ISSUE #71: How to get current tab from outside the TabContext?
- [ ] How to get the currently focused tab from refMap, previously you could use `navigationState.index`?
- [ ] How to listen to the change of the tab? (Connected to above)

If you want to conditionally display content based on the current tab, or you just want to know for some logic which tab is currently focused, you used to be able to work this out from `navigationState`.

PedroBern:
react-navigation support on v3 🎉  

I will update the readme and improve the examples before merging

@andreialecu

```tsx
// example usage

import { createCollapsibleNavigator } from 'react-native-collapsible-tab-view'

export type Params = {
  article: undefined
  albums: undefined
  contacts: undefined
}

const Tabs = createCollapsibleNavigator<Params>()

// you get:
// Tabs.Navigator
// Tabs.Screen
// Tabs.FlatList
// Tabs.ScrollView

<Tabs.Navigator containerRef={containerRef} refMap={refMap} {...props}>
    <Tabs.Screen
      name="article"
      component={ArticleScreen}
      options={{ tabBarLabel: 'Article' }}
    />
    <Tabs.Screen
      name="albums"
      component={AlbumsScreen}
      options={{ tabBarLabel: 'Albums' }}
    />
    <Tabs.Screen
      name="contacts"
      component={ContactsScreen}
      options={{ tabBarLabel: 'Contacts' }}
    />
</Tabs.Navigator>
```

Closes #64 
Closes #71

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,161 +1,189 @@
-# [3.5.0](https://github.com/PedroBern/react-native-collapsible-tab-view/compare/v3.4.0...v3.5.0) (2021-01-27)
-
+## [3.7.1](https://github.com/PedroBern/react-native-collapsible-tab-view/compare/v3.7.0...v3.7.1) (2021-02-01)
 
 ### Bug Fixes
 
-* correct typo "cancelLazyFazeIn" ([2b8b194](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/2b8b1942a84f4fac81f2b5715a005bd288da0421))
-* fix diffClamp tabs sync ([304e8cc](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/304e8ccb14f78ce3eb565699d6e12ec055a8286e))
-* prevent not syncing on tab item press and momentum scrolling ([5c5ff49](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/5c5ff49f5e8ae05a1b8f379b44c1063470fae88c))
+- ensure correct tab label ordering ([1c2e65e](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/1c2e65e0f99fe46c8996f9d189475c4d23dc4ea1)), closes [#77](https://github.com/PedroBern/react-native-collapsible-tab-view/issues/77)
+- handle scroll to top ios ([d96f2e2](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/d96f2e2c35d9bc45c229f4fca243dac2cf12cc10)), closes [#66](https://github.com/PedroBern/react-native-collapsible-tab-view/issues/66)
 
+# [3.7.0](https://github.com/PedroBern/react-native-collapsible-tab-view/compare/v3.6.3...v3.7.0) (2021-01-30)
+
+### Bug Fixes
+
+- add missing tabbar style prop and improve typing ([826966d](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/826966db89141b04c1521f710608686bcc7f5ff8)), closes [#64](https://github.com/PedroBern/react-native-collapsible-tab-view/issues/64)
+- change keyExtactor to use index and to be useCallback ([95be6c2](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/95be6c2ff00bd2ba183126e3e6892e0730e8f0c9))
+- ensure children is not undefined ([bd4f0e2](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/bd4f0e20b66198c8607d1dd8d1a52a1f1a2b2f52)), closes [#68](https://github.com/PedroBern/react-native-collapsible-tab-view/issues/68) [#72](https://github.com/PedroBern/react-native-collapsible-tab-view/issues/72)
 
 ### Features
 
-* add initialTabName prop ([7166592](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/7166592f5258253850722be680ada8197750ef97))
+- add imperactive ref handler ([a6c20fe](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/a6c20feadb0b08b87eb3c98658e31184a039d4b8)), closes [#71](https://github.com/PedroBern/react-native-collapsible-tab-view/issues/71)
+- add onIndexChange prop ([15a516c](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/15a516c21908cf16afd9688c39d10ee34e8e5c35)), closes [#71](https://github.com/PedroBern/react-native-collapsible-tab-view/issues/71)
+- export useCollapsibleStyle hook ([dde28fe](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/dde28fecf0c65ef85dc5c5f54dd26b98fd6e35f0))
+
+## [3.6.3](https://github.com/PedroBern/react-native-collapsible-tab-view/compare/v3.6.2...v3.6.3) (2021-01-30)
+
+### Bug Fixes
+
+- allow flatlist to bounce ([f359f80](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/f359f80f5298686e44116c4ada6a2b8d05c8559f))
+- fix ios without momentum scrolling ([87e3124](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/87e3124eda0903409fd2a3fca802b80ac9251a6f))
+
+### Performance Improvements
+
+- optimize useAnimatedReaction ([9b5b0a9](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/9b5b0a9cf8c1e803c845052449c3f704b1978f4c))
+
+## [3.6.2](https://github.com/PedroBern/react-native-collapsible-tab-view/compare/v3.6.1...v3.6.2) (2021-01-30)
+
+### Bug Fixes
+
+- start with correct calculateNextOffset.value ([fb405f7](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/fb405f7b4980d233930bc06fb31bda875732bf0b))
+
+## [3.6.1](https://github.com/PedroBern/react-native-collapsible-tab-view/compare/v3.6.0...v3.6.1) (2021-01-29)
+
+### Bug Fixes
+
+- prevent diffclamp not syncing on tab press ([1ec9dcd](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/1ec9dcd56386726771d555ce990e7d26fb0c0101))
+
+# [3.6.0](https://github.com/PedroBern/react-native-collapsible-tab-view/compare/v3.5.0...v3.6.0) (2021-01-29)
+
+### Features
+
+- add tab name context, remove name prop from scrollable components ([c5166d2](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/c5166d2a957fb644e039bb357198a06fdbd12323))
+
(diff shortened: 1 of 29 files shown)


andreialecu:
Hey, so I wanted to try this but it seems that when adding it as a git dependency, it fails to build:

```
[96msrc/Navigator/createCollapsibleNavigator.tsx[0m:[93m34[0m:[93m10[0m - [91merror[0m[90m TS2742: [0mThe inferred type of 'createCollapsibleNavigator' cannot be named without a reference to 'react-native/node_modules/@types/react'. This is likely not portable. A type annotation is necessary.

[7m34[0m function createCollapsibleNavigator<T extends ParamListBase>() {
[7m  [0m [91m         ~~~~~~~~~~~~~~~~~~~~~~~~~~[0m

[96msrc/Navigator/createCollapsibleNavigator.tsx[0m:[93m34[0m:[93m10[0m - [91merror[0m[90m TS4058: [0mReturn type of exported function has or is using name 'AnimatedNode' from external module "react-native-reanimated" but cannot be named.

[7m34[0m function createCollapsibleNavigator<T extends ParamListBase>() {
[7m  [0m [91m         ~~~~~~~~~~~~~~~~~~~~~~~~~~[0m


Found 2 errors.
```

The version I was using previously worked via its prepare script, something probably regressed in that area.


```
    "react-native-collapsible-tab-view": "PedroBern/react-native-collapsible-tab-view#feat-react-navigation",
```

andreialecu:
As a quick repro (I'm using yarn 2):

Make sure you have a recent yarn 1 version:
```
npm install -g yarn
```

```
cd `mktemp -d` # switch to a temp dir
yarn init -2 # init on yarn v2
yarn config set nodeLinker node-modules
yarn config set enableInlineBuilds true # makes error easier to see

yarn add react-native-collapsible-tab-view@PedroBern/react-native-collapsible-tab-view#feat-react-navigation
```

andreialecu:
Easier repro:

I cloned the branch this PR is based on. Running `yarn prepack` results in the errors above.

PedroBern:
I found this on StackOverflow, same error with a different package, maybe the second answer help

https://stackoverflow.com/questions/63806168/need-help-fixing-or-suppressing-this-tslint-error-ts2742

I will try to reproduce here

PedroBern:
Same error here 😞  I will investigate

PedroBern:
try now, should work

andreialecu:
You can fix one of the errors by removing `@types/react` from `devDependencies` which shouldn't be needed I think. `@types/react-native` has react types already. In my RN projects I don't add `@types/react`.

The problem with the latest commit is that it loses all types, so code completion won't work

PedroBern:
> You can fix one of the errors by removing @types/react from devDependencies which shouldn't be needed I think. @types/react-native has react types already. In my RN projects I don't add @types/react.

Thanks for the suggestion, I will do it

> The problem with the latest commit is that it loses all types, so code completion won't work

Yes I know, I will fix it

andreialecu:
The other error seems to originate from the `ScrollView` export, aka these two:
```
// Tabs.FlatList
// Tabs.ScrollView
```

I haven't looked too deeply into the changes, but why are they reexported from here?

andreialecu:
To clarify, this diff resolves the build error:

```diff
diff --git a/package.json b/package.json
index ee3aef7..b40c5d1 100644
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "@react-native-community/bob": "^0.16.2",
     "@react-navigation/native": "^5.9.2",
     "@release-it/conventional-changelog": "^1.1.4",
-    "@types/react": "^16.9.44",
     "@types/react-native": "0.63.4",
     "babel-jest": "^26.2.2",
     "babel-preset-react-native": "^4.0.0",
diff --git a/src/Navigator/createCollapsibleNavigator.tsx b/src/Navigator/createCollapsibleNavigator.tsx
index d784e83..4106944 100644
--- a/src/Navigator/createCollapsibleNavigator.tsx
+++ b/src/Navigator/createCollapsibleNavigator.tsx
@@ -32,12 +32,7 @@ import {
 } from './types'
 
 function createCollapsibleNavigator<T extends ParamListBase>() {
-  const {
-    FlatList,
-    ScrollView,
-    Container,
-    useTabsContext,
-  } = createCollapsibleTabs<keyof T>()
+  const { Container, useTabsContext } = createCollapsibleTabs<keyof T>()
 
   const Context = React.createContext<TabOptionsContext<keyof T> | undefined>(
     undefined
@@ -234,8 +229,6 @@ function createCollapsibleNavigator<T extends ParamListBase>() {
   return {
     Navigator,
     Screen,
-    FlatList,
-    ScrollView,
     useTabsContext,
     useTabOptionsContext,
   }
diff --git a/yarn.lock b/yarn.lock
index d6a29fa..3399904 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -2058,14 +2058,6 @@
     "@types/prop-types" "*"
     csstype "^3.0.2"
 
-"@types/react@^16.9.44":
-  version "16.14.2"
-  resolved "https://registry.yarnpkg.com/@types/react/-/react-16.14.2.tgz#85dcc0947d0645349923c04ccef6018a1ab7538c"
-  integrity sha512-BzzcAlyDxXl2nANlabtT4thtvbbnhee8hMmH/CcJrISDBVcJS1iOsP1f0OAgSdGE0MsY9tqcrb9YoZcOFv9dbQ==
-  dependencies:
-    "@types/prop-types" "*"
-    csstype "^3.0.2"
-
 "@types/responselike@*", "@types/responselike@^1.0.0":
   version "1.0.0"
   resolved "https://registry.yarnpkg.com/@types/responselike/-/responselike-1.0.0.tgz#251f4fe7d154d2bad125abe1b429b23afd262e29"
```

But not sure if removing those is correct.

PedroBern:
Yes, I just saw it too, the error comes from scroll view. We need to re-export because they are "connected" to the onScroll of each tab. The `Tabs.ScrollView` knows if the parent component/tab is focused or not, allowing the `Tabs.Container` to track the correct y position (from the currently focused tab) and know how to calculate the offset.

PedroBern:
now it's fixed 👍

andreialecu:
I probably need to open a new issue, but...

I just tried v3 for the first time in Expo on my iPhone, and I'm seeing this:

IMG_BF8E8794E1CC-1

Also the snap example doesn't snap at all, and I cannot tap the tab titles to change pages (they don't do anything, only swiping does)

I'm on this branch, but it doesn't seem related to it. I could try `main` as well.

andreialecu:
Actually it only snaps if momentum scrolling rests within the header. 

But if I manually scroll and lift the finger with the header partly visible it doesn't snap. And the scroll positions of the other tabs are not synced, so they end up like the screenshot above.

PedroBern:
#59 is pointing the same problems 😞  everything is perfect here on my android. Didn't test on iOS so I couldn't tell. Do you have these bugs on android too?

andreialecu:
Didn't test Android. It was quite tricky to get all of these right on iOS on the initial version I contributed to.

In its current state it seems a bit broken on iOS unfortunately :(

There are some other issues as well, such as overscrolling not working any more. Overscrolling is somewhat important to make scrollviews feel native on iOS, otherwise they just abruptly stop at the end, without the nice bounce.

andreialecu:
I'll try to take a look at some of these over the weekend.

PedroBern:
😞  I'm sorry to hear that. My mac mini is arriving in 2 weeks 🎉  so I will be able to investigate further, but as always, all the help is appreciated. Also, if we can't make it better, we can revert to v2...

alexpchin:
@PedroBern I'm sure we can get v3 to work. I think a lot of issues are connected together?

PedroBern:
> @PedroBern I'm sure we can get v3 to work. I think a lot of issues are connected together?

Finally, with #65  or #67, we are getting there 🎉

andreialecu:
I'd like to give this another try, but looks like it needs to be rebased.

PedroBern:
yes sure, also I saw it is broken with diffClamp, don't know why. I need to go away right now and will take some hours to get back, then I resolve all these conflicts.

PedroBern:
@andreialecu it's up to date now. I need to update the example app to have 2 screens, one with react-navigation examples and the other without it. Both screens with the same set of demos/examples.

PedroBern:
@andreialecu [see this note on going back](https://github.com/PedroBern/react-native-collapsible-tab-view/pull/58/files#diff-b2a4de61f1bfc596d553f0c11c6956e74357747f6e92f0982291e00e852d6a41R142)

andreialecu:
Seems this is a bigger breaking change than I anticipated. Will need to go through the example and docs.

There are a bunch of things missing that I noticed at first glance:

```
sceneContainerStyle
```

Screenshot 2021-02-01 at 21 54 06

```
tabBarLabel
```

Screenshot 2021-02-01 at 21 54 24

andreialecu:
Actually all the props of `Navigator` seem to be missing. Including the other one visible in the screenshot (`collapsibleOptions`)

andreialecu:
Ok, I see they're direct properties. Except `animatedValue` which I wasn't able to find yet. 🤔

PedroBern:
It's brand new, has nothing to do with the v1/v2, I have built from the ground...

PedroBern:
There isn't animatedValue. This may be a problem, since we have many shared values now, instead of the animatedValue.

The solution can be making some shared values controlled, just like refs.

andreialecu:
I know that - thought it would keep the API more or less though 🙂 

I'm leaving some more comments here as I go:

It appears that when changing `Animated.FlatList` into `TabsNavigator.FlatList` some problems appear:

```
Animated.FlatList
```
Screenshot 2021-02-01 at 22 23 51

```
TabsNavigator.FlatList
```
Screenshot 2021-02-01 at 22 24 51

`item` gets typed as `unknown`. `ref` doesn't seem to exist.

PedroBern:
> ref doesn't seem to exist.

This ref will be the same as in the refMap, you already have it.

> item gets typed as unknown

Try passing the type explicitly to the flatlist `<TabsNavigator.FlatList<ItemType> ... />`, should work, but it also should work implity 🤔

andreialecu:
> This ref will be the same as in the refMap, you already have it.

Seems it might be a little tricky to pass it down since the `refMap` is defined in the parent, and these are separate individual components. 🤔 

It's likely I'm doing things wrong though, it's late here. Will need to revisit this when I'm less tired. Will take another look tomorrow. 🙂

PedroBern:
> Seems it might be a little tricky to pass it down since the refMap is defined in the parent, and these are separate individual components. 🤔

You can get from the context `useTabsContext` 🙂

andreialecu:
> Try passing the type explicitly to the flatlist `<TabsNavigator.FlatList<ItemType> ... />`, should work, but it also should work implicity 🤔

This didn't seem to work btw. Tried `<any>` without success. I didn't look to deeply into it, but it's possible something is wrong with the types generated by `bob`.

yfunk:
Just out of curiosity, whats the current status on `react-navigation` integration for v4? Is it still something you are planning to add?

PedroBern:
@yfunk not anytime soon, maybe in the future. If you want to work on a PR would be awesome :) As a starting point you can see the [createCollapsibleNavigator](https://github.com/PedroBern/react-native-collapsible-tab-view/blob/feat-react-navigation/src/createCollapsibleTabs.tsx) from this branch, but keep in mind it was built on top of `v3`, so it needs an update.

PR #65, opened 2021-01-30
fix: ios snap behaviour

andreialecu:
On iOS, `onMomentumEnd` is not called unless the user is scrolling with momentum.

This ensures that the snap logic is called properly when momentum scrolling is not in use.

Additionally, it reenables overscroll/bouncing the scroll view.

Fixes #59

PR #67, opened 2021-01-30
Fix ios behaviour (2)

### LINKED ISSUE #59: Jittery behaviour?
Hi @PedroBern, I've finally gotten around to have a look at the new v3 implementation. 

On running the example, I see: `You specified `onScroll` on a <ScrollView> but not `scrollEventThrottle`. You will only receive one event. Using `16` you get all the events but be aware that it may cause frame drops, use a bigger number if you don't need as much precision.`

I don't know if it's related yet, but I'm also seeing quite jittery behaviour. Please see video:

[simulator1.mov.zip](https://github.com/PedroBern/react-native-collapsible-tab-view/files/5893215/simulator1.mov.zip)

I haven't fully investigated this and will add more information if/when I find out more.

PedroBern:
An alternative to #65.

@andreialecu, I didn't test it (I can't), but what you think?

Instead of using setTimeout, I handled everything on the UI, can you run the preview to see if it has the same behavior as your initial fix, please?

Also, having `isGliding.value=true` before `isScrolling.value=false` ensures we sync tabs **after** the snap, [because of here](https://github.com/PedroBern/react-native-collapsible-tab-view/blob/main/src/createCollapsibleTabs.tsx#L581).

Close #59

### DIFF
--- a/src/createCollapsibleTabs.tsx
+++ b/src/createCollapsibleTabs.tsx
@@ -218,7 +218,8 @@ const createCollapsibleTabs = <
             }
           }
         }
-      }
+      },
+      []
     )
 
     const renderItem = React.useCallback(
--- a/src/createCollapsibleTabs.tsx
+++ b/src/createCollapsibleTabs.tsx
@@ -680,7 +680,6 @@ const createCollapsibleTabs = <
       <AnimatedFlatList
         // @ts-ignore
         ref={refMap[name]}
-        bounces={false}
         bouncesZoom={false}
         style={[_style, style]}
         contentContainerStyle={[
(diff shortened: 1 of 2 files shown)


alexpchin:
@PedroBern I think this is cleaner than #65. It seems to work at my end.

andreialecu:
I'm away for a bit so I can't test it right now. Seems you only rewrote the `setTimeout` part?

andreialecu:
If @alexpchin confirms it works, feel free to merge this one instead of mine. We can revisit later if more fixes are needed.

alexpchin:
@andreialecu Your changes have been incorporated too 👍  This PR is on top of the changes you made but re-writes the setTimeout and a couple of tidies.

PedroBern:
rebased, just waiting for the preview for the record and will merge 🎉

PR #69, opened 2021-01-30
feat: small v3 fixes

alexpchin:
feat: ensuring that children is not `undefined`

### DIFF
--- a/src/createCollapsibleTabs.tsx
+++ b/src/createCollapsibleTabs.tsx
@@ -317,6 +317,8 @@ const createCollapsibleTabs = <
       [containerRef, refMap]
     )
 
+    const keyExtractor = React.useCallback((_, index) => index.toString(), [])
+
     return (
       <Context.Provider
         value={{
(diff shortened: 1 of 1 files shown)


PedroBern:
@alexpchin thanks for the PR, but not sure about the need of this. The definition of children is:

```ts
 children: React.ReactElement[]
```

[Here](https://github.com/PedroBern/react-native-collapsible-tab-view/blob/main/src/types.ts#L40).

The children are always an array because I didn't see a use-case where we don't have children, or we have only one child. If children are not provided, there is no screen to be rendered. And if we have only one child, there is no need for tabs. Perhaps I'm missing something?

PedroBern:
What do you think about we throwing an error in case we have only one or zero children?

alexpchin:
So whilst unlikely, you might have a scenario where you only have one tab and even more unlikely, undefined.

For example, if you used tabs for a profile page, but a certain user type only had one tab, this would not be an array. It would be more work to create two pages that dealt differently with the header and tabs.

Regarding the undefined, this is really, really unlikely. However, when developing perhaps we throw an error that is more helpful rather than just a crash?

PedroBern:
Yes, I see your point, I agree. I did some changes on #74, let me know your thought

PedroBern:
Closing in favor of #74

PR #74, opened 2021-01-30
Fix ensure array children

### LINKED ISSUE #68: Ensure that children exists
In `createCollapsibleTabs`, we should ensure that children exists as this can be `undefined`:

Line 88:

```ts
const scrollY = useSharedValue([...new Array(children.length)].map(() => 0));
```

Line 106:

```ts
const [data] = React.useState(
  [...new Array(children.length)].map((_, i) => i)
)
```

### LINKED ISSUE #72: Two children with the same key `undefined` In a Bare React Native App
I'm seeing the error:

```
Warning: Encountered two children with the same key, `undefined`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

PedroBern:
an alternative to #69 

@alexpchin  what you think?

Close #72
Close #68

[documentation](https://reactjs.org/docs/react-api.html#reactchildrentoarray)

```jsx
// add unique key to each child
React.Children.toArray(children)
```

### DIFF
--- a/src/MaterialTabBar/Indicator.tsx
+++ b/src/MaterialTabBar/Indicator.tsx
@@ -18,21 +18,31 @@ const Indicator: React.FC<IndicatorProps> = ({
   const opacity = useSharedValue(fadeIn ? 0 : 1)
 
   const stylez = useAnimatedStyle(() => {
-    return {
-      transform: [
-        {
-          translateX: interpolate(
+    const transform =
+      itemsLayout.length > 1
+        ? [
+            {
+              translateX: interpolate(
+                indexDecimal.value,
+                itemsLayout.map((_, i) => i),
+                itemsLayout.map((v) => v.x)
+              ),
+            },
+          ]
+        : undefined
+
+    const width =
+      itemsLayout.length > 1
+        ? interpolate(
             indexDecimal.value,
             itemsLayout.map((_, i) => i),
-            itemsLayout.map((v) => v.x)
-          ),
-        },
-      ],
-      width: interpolate(
-        indexDecimal.value,
-        itemsLayout.map((_, i) => i),
-        itemsLayout.map((v) => v.width)
-      ),
+            itemsLayout.map((v) => v.width)
+          )
+        : itemsLayout[0].width
+
+    return {
+      transform,
+      width,
       opacity: withTiming(opacity.value),
     }
   }, [indexDecimal, itemsLayout])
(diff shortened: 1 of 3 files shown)


alexpchin:
@PedroBern 

I'm liking this. However, this is still saying that there has to be more than one element in children? 

Another example where there might only be one child is. potentially, is if someone wanted to use this package to render a header with the snap, or diffClamp but without a TabBar - perhaps a Searchbox ontop of results?  If we allow one child, then they can re-use the same page structure in the app as a page with more than one tab?

I personally have a use-case in my app (which I know is unusual) where there are multiple user types but the most basic only has one tab as they don't have enough content. However, because of the way the profile page is structured (like Instagram), the user's profile image and details are rendered above the tabs?

PedroBern:
> I'm liking this. However, this is still saying that there has to be more than one element in children?

No, it accepts 1 or more 👍 

```jsx
if (React.Children.count(children) === 0) {
  throw new Error('CollapsibleTabs must have at least one child.')
}
```

alexpchin:
Have you run the Default example and commented out the everything apart from Articles? (Or just leave any other one).

PedroBern:
yes you are right, it throws an error with only 1 child! I will investigate now

PedroBern:
@alexpchin now it's working the problem was with the indicator component.

alexpchin:
I've just pulled the latest version. However, I currrently see:

Screenshot 2021-01-30 at 21 52 37

With the type error being: 

```
(property) Container: React.FC<CollapsibleProps<TabNames, MaterialTabBarProps<TabNames>>>
This JSX tag's 'children' prop expects type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, any>)>[] | ... 6 more ... | (ReactElement<...>[] & ReactPortal)' which requires multiple children, but only a single child was provided.ts(2745)
```

And: 

Screenshot 2021-01-30 at 21 57 14

Is it working for you?

PedroBern:
here is working, but there is this typescript error, I will fix now

WhatsApp Image 2021-01-30 at 19 38 11

PedroBern:
I was having this error before, maybe you are not with the latest version, I rebased and forced push

PedroBern:
@alexpchin just rebased and for push again

PedroBern:
I merged, if you see any errors, let me know

PR #76, opened 2021-01-31
fix: handle scroll to top ios

### LINKED ISSUE #66: Scroll on Header not scrolling other tabs (iOS)
### Current behavior

In the scroll on header example, if I go to contacts and scroll to the end, then tap the status bar, it correctly scrolls the current tab to the top. 

However, all the other tabs are not synced.

### Expected behaviour

All tabs should be scrolled to the top.

### Screenshots (if applicable)

IMG_E4F8296E5D88-1

andreialecu:
Fixes #66 

Open for discussion. I didn't notice any breakage from doing it this way.

The idea behind it is that if we detect `onScroll` being called, then we set `isScrolling` to `1`. Once `onScroll` stops firing then it is set back to `0`.

This has the benefit and side effect that other types of scroll besides those initiated by `onBeginDrag` are handled by default (such as programatic scrolls)

### DIFF
--- a/src/createCollapsibleTabs.tsx
+++ b/src/createCollapsibleTabs.tsx
@@ -102,7 +102,7 @@ const createCollapsibleTabs = <T extends ParamList>() => {
       const [headerHeight, setHeaderHeight] = React.useState<
         number | undefined
       >(initialHeaderHeight)
-      const isScrolling = useSharedValue(false)
+      const isScrolling = useSharedValue(0)
       const scrollYCurrent = useSharedValue(0)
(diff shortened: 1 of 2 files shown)


andreialecu:
This also differs from the current implementation. With this change, `isScrolling` will only be set to a falsy value after the momentum scroll completely finishes (as opposed to immediately after the user lifts the finger).

However, I tested everything and I didn't notice anything breaking.

PedroBern:
Genius solution 👍  Everything is working on android.

PR #79, opened 2021-01-31
feat: showing issue with scrollable-tabs

alexpchin:
Issue with scrollable tabs.


PedroBern:
@alexpchin Don't know how the order got wrong in the first place, but this will ensure the correct order anyway 👍

PR #80, opened 2021-02-01
feat: commenting back in centered list empty but using headerHeight from useTabsContext

alexpchin:


PedroBern:
@alexpchin it gets broken if we add diffclamp to this example

PR #83, opened 2021-02-01
feat: allow tab bar customization

### LINKED ISSUE #82: TabBar Item Label style required
### Feature request
Please provide an option for changing the Tab Bar Item Label Style. 
In #64, you provided an option for changing the tab bar background, that is great. 

But need an option to change the existing label style without using custom 'TabItemComponent' as you mentioned in #70.

Waiting for your valuable reply.

tcorreiaubi:
Hey, this is based mostly on the approach that react-native-tab-view takes about styling their tab bar, the idea is to allow customization without having to refactor/copy the existent bar.

I had to increase the package version because of the following error coming from interpolateColor
https://github.com/software-mansion/react-native-reanimated/issues/1401

Let me know if this changes make sense

Closes #82

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,21 +353,65 @@ You can use this to get the progessViewOffset and pass to the refresh control of
 
 ### MaterialTabBar
 
+Basic usage looks like this:
+```tsx
+import {
+  RefComponent,
+  ContainerRef,
+  TabBarProps,
+} from 'react-native-collapsible-tab-view'
+import { useAnimatedRef } from 'react-native-reanimated'
+type MyTabs = 'article' | 'contacts' | 'albums'
+const MyHeader: React.FC<TabBarProps<MyTabs>> = (props) => {...}
+const Example: React.FC<Props> = () => {
+  const containerRef = useAnimatedRef<ContainerRef>();
+  const tab0Ref = useAnimatedRef<RefComponent>();
+  const tab1Ref = useAnimatedRef<RefComponent>();
+  const [refMap] = React.useState({
+    tab0: tab0Ref,
+    tab1: tab1Ref,
+  });
+  return (
+    <Tabs.Container
+      containerRef={containerRef}
+      HeaderComponent={MyHeader}
+      headerHeight={HEADER_HEIGHT} // optional
+      refMap={refMap}
+      TabBarComponent={(props) => (
+        <MaterialTabBar
+          {...props}
+          activeColor="red"
+          inactiveColor="yellow"
+          labelStyle={{ fontSize: 14 }}
+        />
+      )}
+    >
+      {components returning Tabs.ScrollView || Tabs.FlatList}
+    </Tabs.Container>
+  );
+};
+```
+
 #### Props
 
-|name|type|default|
-|:----:|:----:|:----:|
-|TabItemComponent|`((props: MaterialTabItemProps<any>) => ReactElement<any, string \| ((props: any) => ReactElement<any, string \| ... \| (new (props: any) => Component<any, any, any>)> \| null) \| (new (props: any) => Component<...>)>) \| undefined`|`null`|
-|containerRef|`RefObject<ContainerRef>`||
-|focusedTab|`SharedValue<any>`||
-|getLabelText|`((name: any) => string) \| undefined`|`(name) => name.toUpperCase()`|
-|index|`SharedValue<number>`||
-|indexDecimal|`SharedValue<number>`||
-|indicatorStyle|`StyleProp<AnimateStyle<ViewStyle>>`||
-|onTabPress|`(name: any) => void`||
-|refMap|`Record<any, Ref>`||
-|scrollEnabled|`boolean \| undefined`|`false`|
-|style|`StyleProp<ViewStyle>`||
+|name|type|default|description|
+|:----:|:----:|:----:|:----:|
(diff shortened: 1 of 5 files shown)


PedroBern:
hi @tcorreiaubi thanks for the PR, the changes do make sense. However, expo is officially supporting only rc.0. The rc.1 and rc2 have native code changes that expo didn't follow up yet.

So we need to avoid using `intepolateColor` for the moment 😞 .

You can keep it open and wait for official support or you can come up with a different solution (Probably interpolating the opacity as the current implementation).

tcorreiaubi:
@PedroBern I was actually just looking at it, the interpolateColor behavior ended up looking really not good as expected I guess a simple condition rendering either 1 color or another will do it for the moment, I will update the pull request

tcorreiaubi:
@PedroBern fixed, I'm also working in implementing a minimum collapsible header height, will make the pull request tomorrow.
And after that if I can get sometime to work on it, I will help on the testing, I believe it makes sense to we can easily add more functionality

alexpchin:
@tcorreiaubi @PedroBern I was actually working on pretty much the same PR today... In my project, I have implemented `intepolateColor` (I'm not on expo) but then I realised also I needed a `color` value to pass to my focused custom TabItemComponent because I am using FontAwesome and custom badges for notifications. (FontAwesomeIcons do not work with react-native-reanimated. https://github.com/FortAwesome/react-native-fontawesome/issues/53)

I've not solved it yet...

src/MaterialTabBar/TabItem.tsx
```diff
@@ -32,6 +36,7 @@ const TabItem: React.FC<MaterialTabItemProps<any>> = ({
         [inactiveOpacity, 1, inactiveOpacity],
         Animated.Extrapolate.CLAMP
       ),
+      color: indexDecimal.value === index ? activeColor : inactiveColor,
```
PedroBern, line 39: replace by:

```tsx
color: Math.abs(index - indexDecimal.value) < 0.5 ? activeColor : inactiveColor,
```

src/MaterialTabBar/TabItem.tsx
```diff
@@ -19,6 +21,8 @@ const TabItem: React.FC<MaterialTabItemProps<any>> = ({
   label,
   style,
   labelStyle,
+  activeColor = DEFAULT_ACTIVE_COLOR,
+  inactiveColor = DEFAULT_INACTIVE_COLOR,
```
PedroBern, line 25: As we are already animating the opacity, we can have only one default color. Please keep both black.

PedroBern (changes requested):
Everything looks good, just some minor changes

PedroBern:
> I finally solved it by using '@react-native-community/masked-view' with an animated background under the icon.

@alexpchin clever solution.

tcorreiaubi:
@PedroBern Fixed, I think it can be merged

PedroBern:
@tcorreiaubi nice work 👍 

Need to solve some conflicts, you can copy the example you wrote in the readme and add it as a docstring to the tabbar.  Will need to remove the `/* */` inside the children.

And delete the prop table, it will be auto-generated.

Then, run `yarn docs`

tcorreiaubi:
@PedroBern Done, let me know if something is missing

README.md
```diff
@@ -379,18 +423,20 @@ Any additional props are passed to the pressable component.
...
 |name|`any`|||
 |onLayout|`(((event: LayoutChangeEvent) => void) & ((event: LayoutChangeEvent) => void)) \| undefined`||Invoked on mount and layout changes with {nativeEvent: { layout: {x, y, width, height}}}.|
 |onPress|`(name: any) => void`|||
 |pressColor|`string \| undefined`|`#DDDDDD`||
 |pressOpacity|`number \| undefined`|`Platform.OS === 'ios' ? 0.2 : 1`||
 |scrollEnabled|`boolean \| undefined`|||
-|style|`ViewStyle \| (ViewStyle & false) \| (ViewStyle & number & { __registeredStyleBrand: ViewStyle; }) \| (ViewStyle & RecursiveArray<false \| ViewStyle \| RegisteredStyle<...> \| null \| undefined>) \| (ViewStyle & ((state: Readonly<...>) => StyleProp<...>)) \| undefined`||Either view styles or a function that receives a boolean reflecting whether the component is currently pressed and returns view styles.|
+|style|`false \| ViewStyle \| RegisteredStyle<ViewStyle> \| RecursiveArray<false \| ViewStyle \| RegisteredStyle<ViewStyle> \| null \| undefined> \| ... 15 more ... \| undefined`||Either view styles or a function that receives a boolean reflecting whether the component is currently pressed and returns view styles.|
```
andreialecu, line 439: This looks odd. Might be an issue with doc generation though.

PedroBern: The docs script takes the type definition, in the `types.ts` it is `style?: StyleProp<ViewStyle>`, which resolves to this awkward definition. Works in most cases...

Here we don't need it, since the style is already defined in the PressableProps (it is defined twice here). Was my mistake to add it. Also there is an unused prop `ItemElement` that I need to remove from the types (also my mistake). @tcorreiaubi can you please remove both of them from the `types.ts` and then run the docs script?

tcorreiaubi: Hey @PedroBern style?: StyleProp<ViewStyle> is required here since the PressableProps omit the style from the viewprops 

```typescript
export interface PressableProps extends AccessibilityProps, Omit<ViewProps, 'style' | 'hitSlop'> {
```

PedroBern: You are right! About the weird types in the prop tables, I'm going to solve this later (another day)

PedroBern (approved):

PR #85, opened 2021-02-02
feat: add support for collpased header height

### LINKED ISSUE #5: Define a minimum header hight
Hi, love this plugin well written. 
I'm trying to make a header with several animated objects in it that fade out but some will stay. So far i have not figured out where the final offset is calculated to keep some of the header without closing it compleately

tcorreiaubi:
Hey @PedroBern, this allows to configure a height for the header when collapsed, in other words, the header will be considered collapsed once we reach, header height - collapsed height, I tested all the examples and they seem to be working properly, so I believe I got all the right variables in their place, but if you can give a new look, since you know the project/behavior better than me

Closes #5

### DIFF
--- a/src/createCollapsibleTabs.tsx
+++ b/src/createCollapsibleTabs.tsx
@@ -155,6 +155,7 @@ const createCollapsibleTabs = <T extends ParamList>() => {
         initialTabName,
         containerRef,
         headerHeight: initialHeaderHeight,
+        minHeaderHeight = 0,
         tabBarHeight: initialTabBarHeight = TABBAR_HEIGHT,
         snapEnabled = false,
         diffClampEnabled = false,
@@ -213,6 +214,9 @@ const createCollapsibleTabs = <T extends ParamList>() => {
       const isGliding = useSharedValue(false)
       const endDrag = useSharedValue(0)
       const calculateNextOffset = useSharedValue(index.value)
+      const headerScrollDistance = useDerivedValue(() => {
+        return headerHeight !== undefined ? headerHeight - minHeaderHeight : 0
+      }, [headerHeight, minHeaderHeight])
 
       const getItemLayout = React.useCallback(
         (_: unknown, index: number) => ({
@@ -313,9 +317,13 @@ const createCollapsibleTabs = <T extends ParamList>() => {
               accDiffClamp.value = 0
             } else {
               const nextValue = accDiffClamp.value + delta
+
               if (delta > 0) {
                 // scrolling down
-                accDiffClamp.value = Math.min(headerHeight || 0, nextValue)
+                accDiffClamp.value = Math.min(
+                  headerScrollDistance.value,
+                  nextValue
+                )
               } else if (delta < 0) {
                 // scrolling up
                 accDiffClamp.value = Math.max(0, nextValue)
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -46,6 +46,7 @@ export type CollapsibleProps<T extends ParamList> = {
   initialTabName?: T
   containerRef: React.RefObject<ContainerRef>
   headerHeight?: number
+  collapsedHeaderHeight?: number
```
PedroBern, line 49: Nice work 👍  I tested and everything seems to be working. Just one thing, I think `minHeaderHeight` makes more sense instead of `collapsedHeaderHeight` since we are passing the min value of the header. Can you make this little change, please? Also, it needs to be rebased, now the documentation is auto-generated from the types, see the latest readme.

src/createCollapsibleTabs.tsx
```diff
@@ -232,15 +235,18 @@ const createCollapsibleTabs = <T extends ParamList>() => {
               const nextValue = accDiffClamp.value + delta
               if (delta > 0) {
                 // scrolling down
-                accDiffClamp.value = Math.min(headerHeight || 0, nextValue)
```
PedroBern, line 235: Did you tested the condition

```
if (accScrollY.value <= 0) {
  // handle overscroll on ios, when being dragged beyond the top border
  accDiffClamp.value = 0
```

Some lines before these changes?

Because I can't test on iOS to see how it is handling the overscroll.

src/createCollapsibleTabs.tsx
```diff
@@ -130,6 +131,8 @@ const createCollapsibleTabs = <T extends ParamList>() => {
       const isGliding = useSharedValue(false)
       const endDrag = useSharedValue(0)
       const calculateNextOffset = useSharedValue(index.value)
+      const headerScrollDistance =
```
PedroBern, line 134: This will be recomputed on every rerender, shouldn't we use `useSharedValue` or `useDerivedValue` or `useState`? What you think?

PedroBern: Maybe useDerivedValue is the way to go.

tcorreiaubi: Yeah, makes complete sense, my mistake, fixed

tcorreiaubi:
@PedroBern Related to the snapThreshold, do you think it should be applied to the headerHeight or the headerScrollDistance(headerHeight - minHeaderHeight)? at the moment it's applied to the headerScrollDistance

src/createCollapsibleTabs.tsx
tcorreiaubi: I wasn't able to validate it, also I'm not aware of what should be the behavior, i'm not an iOS user, but if you can explain the behavior I can validate/record it on a simulator

PedroBern:
> @PedroBern Related to the snapThreshold, do you think it should be applied to the headerHeight or the headerScrollDistance(headerHeight - minHeaderHeight)? at the moment it's applied to the headerScrollDistance

headerScrollDistance is correct 👍

src/createCollapsibleTabs.tsx
PedroBern: 😞  I can't test it. This is the behavior: https://medium.com/@wcandillon/ios-bounce-list-effect-with-react-native-5102e3a83999

tcorreiaubi: Just tested it, I believe the behavior is correct, but if someone else can also validate it would be great

PedroBern: Let's merge, if there is an issue then, we fix it.

PR #87, opened 2021-02-03
test: add test setup

### LINKED ISSUE #81: Automated tests
### Feature request

It would be great if a test suite could verify that everything is working properly. The interactions are pretty finicky and things can break in odd places by doing minor changes in others.

This would make the library much more maintainable and easier to refactor in the future, while giving more peace of mind to users that upgrades are safe.

Should be relatively easy to add gray box e2e testing using:
https://github.com/wix/Detox

Tests could be added to initiate scrolls and touches, then assert whether things are positioned correctly.

Another possibility is to convert the Example app to use Storybook:
https://storybook.js.org/
https://www.learnstorybook.com/intro-to-storybook/react-native/en/get-started/

It would then be possible to use loki to do screenshot diff testing for the stories:
https://github.com/oblador/loki

PedroBern:
Closes #81 

Tests are not working, because it's mocking the reanimated, which is essential to exist in the tests 🤔

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  presets: ['module:metro-react-native-babel-preset'],
+}
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,10 @@
+module.exports = {
+  preset: 'react-native',
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
+  modulePathIgnorePatterns: [
+    '<rootDir>/example/node_modules',
+    '<rootDir>/lib/',
+  ],
+  setupFiles: ['<rootDir>/jest/setup.js'],
+  setupFilesAfterEnv: ['@testing-library/jest-native/extend-expect'],
+}
--- a/jest/setup.js
+++ b/jest/setup.js
@@ -1,0 +1,14 @@
+import 'react-native-gesture-handler/jestSetup'
+
+jest.mock('react-native-reanimated', () => {
+  const Reanimated = require('react-native-reanimated/mock')
+
+  // The mock for `call` immediately calls the callback which is incorrect
+  // So we override it with a no-op
+  Reanimated.default.call = () => {}
+
+  return Reanimated
+})
+
+// Silence the warning: Animated: `useNativeDriver` is not supported because the native animated module is missing
+jest.mock('react-native/Libraries/Animated/src/NativeAnimatedHelper')
--- a/package.json
+++ b/package.json
@@ -45,12 +45,17 @@
   },
   "homepage": "https://github.com/PedroBern/react-native-collapsible-tab-view#readme",
   "devDependencies": {
+    "@babel/core": "^7.0.0",
     "@commitlint/config-conventional": "^9.1.1",
     "@microsoft/tsdoc": "^0.13.0",
     "@react-native-community/bob": "^0.16.2",
     "@release-it/conventional-changelog": "^1.1.4",
-    "@types/react": "^16.9.44",
-    "@types/react-native": "0.63.4",
+    "@testing-library/jest-native": "^3.4.3",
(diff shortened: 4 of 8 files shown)


andreialecu:
Is this the right approach though? It wouldn't test actual on-device interaction which is what we'd be after here. 🤔  [Detox](https://github.com/wix/detox) on the other hand would do it.

PedroBern:
> Is this the right approach though?

No. Jest needs Reanimated to be mocked. We need Reanimated to test the app.

> Detox on the other hand would do it.

Yes we need to add detox or appium. I have no experience with any of them, but detox seems a better option since it's RN focused.

What would be the best approach? Some consideration points, please correct me if I'm wrong:

- Adding detox will need to build the app before the tests.
- Tests can be added to the example folder, and leverage the expo app to make the builds
- Adding tests into the example folder will need to resolve the components outside the node_modules (../src)
- CircleCI offers a little macOS free tier to run an iOS simulator (I already requested to my account), not sure if the free tier is enough
- On the other hand, testing on android is cheaper (Linux)
- Question: After running the app on simulator, we can use jest to test without mocking Reanimated?

What am I missing?

andreialecu:
I have some experience with Detox. It's not hard to add and work with.

In my experience, CircleCI cannot run Android emulators, just iOS: https://support.circleci.com/hc/en-us/articles/360000028928-Testing-with-the-Android-emulator-on-CircleCI

I do have a project that runs Detox on CircleCI with iOS though, and it runs properly.

When I did my initial research on Appium vs Detox, I found a lot of issues with Appium which were a non-starter. I don't remember what they were exactly, but I don't recommend it.

Detox can just be added as a script to package.json:

```
    "test": "detox test -c ios.sim.debug main.spec.js",
```

Running `yarn test` then spawns the iOS simulator, runs the test, and closes it. It's pretty simple.

andreialecu:
Note that the above comment refers to using detox with vanilla react-native (not expo).

Expo may make things easier: https://blog.expo.io/testing-expo-apps-with-detox-and-react-native-testing-library-7fbdbb82ac87

PedroBern:
> Expo may make things easier: https://blog.expo.io/testing-expo-apps-with-detox-and-react-native-testing-library-7fbdbb82ac87

Nice article. Indeed looks pretty straightforward. I will dig more into this once I'm able to run on iOS, so I can test locally, then try on circleCI. Up to 2 more weeks until my m1 arrives 🎉   If you need these tests to use with confidence on your project and want to take ahead it's ok for me, I will follow closely.

andreialecu:
I think you can test locally on Android with Detox as well.

PedroBern:
> I think you can test locally on Android with Detox as well.

I do not want to be boring, but my computer is bad, I am waiting for the upgrade to run any simulator. I have developed only on the real devices.

andreialecu:
Ah, makes sense. I see.

andreialecu:
Note that it may be possible to run it on a real Android device: https://github.com/wix/Detox/issues/968

PedroBern:
That's nice, I will look into it later :)

alexpchin:
@PedroBern @andreialecu I have some experience with detox FYI. I would choose detox over Apium personally. I can help with this once I have fixed the current issues and release the latest build of the app I am working on.

PedroBern:
@alexpchin thanks, all help is welcome 🙌

shail-spreadd:
@PedroBern i am getting error while writing test cases using jest 
Can you please help me? my timeline is on monday and i need to complete the test cases for this package

``` The above error occurred in the <ForwardRef> component:
    
    node_modules/react-native-collapsible-tab-view/lib/commonjs/Container.tsx:65:9
    ```
    
jest.setup.js

    ```global.__reanimatedWorkletInit = () => {};
jest.mock('react-native-reanimated', () =>
  require('react-native-reanimated/mock')
);```

andreialecu:
@shail-spreadd please open a new issue.

PR #90, opened 2021-02-03
refactor: polish the api with custom hooks

andreialecu:
The api seemed a bit cumbersome with all the animated refs being passed around, so I added some custom hooks.

PR is not yet complete: The documentation needs to be updated, but everything seems to work.

### DIFF
--- a/src/hooks.tsx
+++ b/src/hooks.tsx
@@ -1,0 +1,23 @@
+import { useState } from 'react'
+import { ContainerRef, RefComponent } from 'react-native-collapsible-tab-view'
+import { useAnimatedRef } from 'react-native-reanimated'
+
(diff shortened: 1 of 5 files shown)


src/hooks.tsx
```diff
@@ -0,0 +1,23 @@
+import { useState } from 'react'
+import { ContainerRef, RefComponent } from 'react-native-collapsible-tab-view'
```
PedroBern, line 2: Isn't keep relative imports better? `./types` To avoid any circular import. If I'm missing something let me know.

andreialecu: Indeed. Looks like the auto import fix messed this up.

However. I'd prefer #93 to this.

PR #91, opened 2021-02-03
refactor: remove need to use any hooks

andreialecu:
This is an alternative to #90 that simplifies things further.

`containerRef` and `refMap` are no longer needed, the tab names are passed to `createCollapsibleTabs` instead.

I'm not sure if this limits things in any way.

Open for discussion. 

With this PR only `const { useTabsContext, ...Tabs } = createCollapsibleTabs(TabNames)` is needed.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -27,3 +27,5 @@ export type {
 export { default as createCollapsibleTabs } from './createCollapsibleTabs'
 
 export { default as MaterialTabBar } from './MaterialTabBar'
+
+export * from './hooks'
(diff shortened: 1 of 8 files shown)


andreialecu:
This would be a breaking change btw.

src/index.tsx
```diff
@@ -27,3 +27,5 @@ export type {
 export { default as createCollapsibleTabs } from './createCollapsibleTabs'
 
 export { default as MaterialTabBar } from './MaterialTabBar'
+
+export * from './hooks'
```
andreialecu, line 31: Probably no need to export the hooks with this implementation.

andreialecu:
I like #93 better btw.

PedroBern:
> I like #93 better btw.

This is great too, but I like #93 better too!

Issue #88, opened 2021-02-03
Tab Screen is showing up blank


danielfein:
### Current behavior

Currently I have 4 screens and set them each to have a width of the window. Despite this, when swiping from one to the other the next screen is visible but then when you get to the point of the index changing, it disappears. If you click to another tab the same thing happens, it's just blank.

### Expected behaviour

I would expect to see the relevant screen on both swipe and when each tab is selected.

### Code sample

```
const { useTabsContext, ...Tabs } = createCollapsibleTabs()

const Page = () => {
    const [index, setIndex] = useState(0);

    const containerRef = useAnimatedRef()
    const postsTabRef = useAnimatedRef()
    const detailsTabRef = useAnimatedRef()
    const itemsTabRef = useAnimatedRef()
    const similarTabRef = useAnimatedRef()
    let refs = {
                    Posts: postsTabRef,
                    Details: detailsTabRef,
                    Items: itemsTabRef,
                    Similar: similarTabRef 
                }
    const [refMap, setRefMap] = React.useState(refs)


    return (<Tabs.Container
                containerRef={containerRef}
                headerHeight={headerHeight} // optional
                refMap={refMap}
                scrollEnabled = {true}
                onIndexChange={setIndex}
                >
                <Tabs.ScrollView style={{ width: windowWidth, height: 600, backgroundColor: 'red'}} contentContainerStyle={{ height: "100%" }}>
                    <View><Text>Hello</Text></View>
                </Tabs.ScrollView>
                <Tabs.ScrollView style={{ width: windowWidth, height: 600, backgroundColor: 'orange'}} />
                <Tabs.ScrollView style={{ width: windowWidth, backgroundColor: 'yellow'}} />
                <Tabs.ScrollView style={{ width: windowWidth, backgroundColor: 'green'}} />

                </Tabs.Container>
        )
}
```

### Screenshots (if applicable)

https://user-images.githubusercontent.com/1616865/106750608-0bd80a00-65f6-11eb-85e4-738a806468fc.mov

### What have you tried

I originally had the 4 `<Tabs.ScrollView />` components as the ones that are actually use in my app. I've minimized down to the bare elements to try and see where I may be causing my error but even with this basic layout it isn't working.

I've tried alternative widths and flex: 1, etc. I can't figure out where the issue is.

danielfein:
Interesting, by removing onIndexChange this worked. Any idea why?

PedroBern:
[...]

Issue #70, opened 2021-01-30
Make `refMap` more customizable


alexpchin:
- [ ] It would be good to be able to pass a custom Component or custom Text to TabBarItem label
- [ ] The keys may be sorted alphabetically by prettier if you use: `sort-keys-fix/sort-keys-fix` or `sort-keys` plugins
- [ ] Unusual (I know) but an empty object would error in the current structure. We should perhaps handle this?

Currently, the `refMap` uses the keys to create the labels. 

Line 150 in src/MaterialTabBar/TabBar.tsx

```ts
getLabelText = (name) => name.toUpperCase()

.
.
.

{Object.keys(refMap).map((name, i) => {
        return (
          <TabItemComponent
            key={name}
            index={i}
            name={name}
            label={getLabelText(name)}
            onPress={onTabPress}
            onLayout={scrollEnabled ? onTabItemLayout : undefined}
            scrollEnabled={scrollEnabled}
            indexDecimal={indexDecimal}
            {...tabItemProps}
          />
        )
      })}
```

However, if we changed the structure of refMap, it could take a label or component/icon to might allow for more customisation.

Something like:

```ts
const [refMap] = React.useState({
  screen1: { ref: ref, label: () => <FontAwesomeIcon icon={'images'} size={22} /> },
  screen2: { ref: ref, label: () => <Text>With Space</Text> }
})
```

PedroBern:
>  It would be good to be able to pass a custom Component or custom Text to TabBarItem label

We can achieve this with something like this (not documented yet):

```jsx
const MyCustomTabItem = ({ name }) => { return <Icon ... />}
...
<Tabs.Container 
   ...
   TabBarComponent={props => <MaterialTabBar TabItemComponent={MyCustomTabItem} />}
/>
```

Also, we probably will add something in #58 or later.

> The keys may be sorted alphabetically by prettier if you use: sort-keys-fix/sort-keys-fix or sort-keys plugins

It's a problem, maybe changing refMap to be an array (?) We would need to review all the code. Right now everything depends on this order. Maybe we can just add some info in the docs saying it.

> Unusual (I know) but an empty object would error in the current structure. We should perhaps handle this?

What would be the use case with an empty refMap? There would be no tabs/screens

PedroBern:
closes by #93

PR #93, opened 2021-02-03
Version 4.0 with api refactoring and dynamic tabs

andreialecu:
Third alternative to #90 or #91 😆 

Introduces a `<Tabs.Tab />` component for defining the tabs.

```tsx
      <Tabs.Container ref={ref} headerHeight={HEADER_HEIGHT} {...props}>
        <Tabs.Tab name="article">
          <Article />
        </Tabs.Tab>
        <Tabs.Tab name="albums">
          <Albums />
        </Tabs.Tab>
        <Tabs.Tab name="contacts">
          <Contacts emptyContacts={emptyContacts} />
        </Tabs.Tab>
      </Tabs.Container>
```

I think I like this the best personally. Seems most idiomatic with React practices in general.

Another benefit is that there's no more need to define the tabs twice (and to make sure the order matches).


andreialecu:
This may have the benefit that it's easy to add/remove tabs at runtime dynamically. I'm not sure if this part works, because I haven't tested it. But if it doesn't, it should be fixable.

andreialecu:
I pushed another minor commit but it seems that GitHub is having problems because it's not displaying it: https://www.githubstatus.com/

andreialecu:
Also, this is a breaking change so it would need to bump the package to `4.0`.

PedroBern (approved):
These changes are so awesome 🎉  🚀  Makes the API much more idiomatic, I really like it.

Just one concern, how do we get the ref from a parent component? Maybe a simple solution is to somehow map the refMap to an object inside the useImperativeHandle. So we can access like this: `ref.current.article.current...` or `ref.current.tabs.article.current...`.

Would be awesome if it were possible to pass props to the `Tab` and send it to `TabBarComponent` or `HeaderComponent`. Just like react-navigation. But that's another story...

PedroBern:
I'm thinking about how could we achieve an expo preview for PRs other than mine. The github secrets can't be shared between forks, so seems there is no right solution for this. Would be great if it were possible. Right now I need to fetch every PR and test locally.

PedroBern:
@andreialecu we can add something like this:

```tsx
// utility function
const extractProps = (children: React.ReactElement[], names: T) => {
  const props = {}
  React.Children.forEach(children, (element, index) => {
    props[names[index]] = element.props
  })
  return props
}

// inside the Container:

// extract the props passed to each Tab component and map to the tab names
const tabOptions = React.useRef(extractProps(children, names))

// pass the tab options to the TabBar
<TabBarComponent options={tabOptions} />
```

And then use like this:

```jsx
      <Tabs.Container ref={ref} headerHeight={HEADER_HEIGHT} {...props}>

        // direct props
        <Tabs.Tab name="article" tabLabel="Article" icon={...}>
          <Article />
        </Tabs.Tab>

        // or options
        <Tabs.Tab name="albums" options={{ icon: ..., label: .... }}>
          <Albums />
        </Tabs.Tab>

      </Tabs.Container>
```

PedroBern:
I have just added the options 🚀  Please review and make necessary changes if needed, including the naming conventions.

src/Tab.tsx
```diff
@@ -2,9 +2,9 @@ import React from 'react'
 
 import { ParamList, TabOptions } from './types'
 
-export interface TabProps<T extends ParamList> {
+export type TabProps<T extends ParamList> = {
```
PedroBern, line 5: I changed to `type` instead of `interface` because the docs scrip doesn't like interfaces.

src/hooks.tsx
```diff
@@ -0,0 +1,44 @@
...
+    (refs, tabId) => ({ ...refs, [tabId]: useTabRef() }),
+    {} as Record<ParamList, React.RefObject<RefComponent>>
+  )
+
+  const [refMap] = useState(refs)
+  return refMap
+}
+
+export const useTabOptions = <T extends ParamList>(
+  children: React.ReactElement<TabProps<T>>
```
PedroBern, line 29: this type can probably be narrowed down.

src/types.ts
```diff
@@ -152,3 +153,12 @@ export type CollapsibleStyle = {
   }
   progressViewOffset: number
 }
+
+export type TabOptions = {
+  label?: string
+}
+
+export type FinalTabOptions<T extends ParamList> = Record<
```
PedroBern, line 161: Do you have a better name for this? FinalTabOptions is awkward.

andreialecu:
I'm working on this at the moment btw. Will report back in a few.

andreialecu:
Latest commit fixes #88 and adds dynamic tabs so it also fixes #94.

`refMap` no longer exists so it should also close #70 

I notice a remaining bug with `Lazy` mode which also exists on `main` (and I noticed it previously). 

Sometimes, randomly, a page will open with opacity `0` and won't fade in. Only occurs with lazy so there's something in it affected by some race condition.

andreialecu:
IMG_874E6D02897B-1

Here's what I mean on the lazy issue. It very rarely happens. And it only happens to one specific tab, scrolling to others make them work.

I attempted a fix in the latest commit but it wasn't it.

andreialecu:
Actually I can reproduce this randomly on the Scrollable Tabs example as well, so it's not related to lazy mode. I can also reproduce it on `main`.

andreialecu:
Something is not right with `ref` handling for lazy routes, working on it.

PedroBern:
> Something is not right with ref handling for lazy routes, working on it.

If you add:

```diff
  function setRef<T extends RefComponent>(key: TabName) {
++  console.log('creating ref:', key)
    // eslint-disable-next-line react-hooks/rules-of-hooks
    const ref = useAnimatedRef<T>()
    map[key] = ref
    return ref
  }
```

You see it's running 3 times for each tab on every example. For the lazy ones, it runs one additional time after visiting the route. The 3 times are from the onLayout callbacks (container, tabbar, header), passing the header height should reduce to 2 rerenders (somehow on my android I pass 250,  but always result in 250.58 after the layout, nvm)

The additional time for the lazy tabs is due to the `Tabs.ScrollView` inside `Tabs.Lazy`. The fake scrollview call 3 times, and then the actual component calls one more time when mounted. I added it to have the ref from `refMap` connected to a scrollable component even when there are no children, it would help to keep the correct scrollYs and offsets.

Now, as the refs are dynamically created, maybe there is an issue here, not sure yet.

andreialecu:
I was working on some fixes, it seems good now. 

Lazy loading seems to work as well.

Previously there was a bug due to attempting to scroll the scrollview before the children were rendered. So if you were slightly scrolled while navigating to a new tab, it wouldn't position the new tab properly.

andreialecu:
https://user-images.githubusercontent.com/697707/106931598-6caf3180-671f-11eb-98c6-84cbd1bb1d0a.mov

There's a bug here due to rerendering, which is also related to the `onIndexChange` problem in #88.

The tabbar rerenders, loses its scroll position, and always starts scrolling from the start while swiping to other tabs. There's also a flicker of the indicator which is slightly annoying.

If you want to take a look at it it would be great. I'm going to be away for a bit.

andreialecu:
Seems I also broke the `Remove Tab` functionality somehow in recent commits (from the dynamic example)

PedroBern:
Support dynamic tabs will need a tab bar rework. It was initially designed for static tabs.

The annoying flicker from the indicator is due to:

```tsx
...
})}
   {itemsLayout.length === nTabs && (
      <Indicator
        ...
```

As we get new tabs length, it unmounts and mounts.

> There's a bug here due to rerendering, which is also related to the onIndexChange problem in #88.

The Container really dislikes rerendering, that's why I wrapped it with `useMemo` both in the #58 and in the `onIndexChange` example.

It's like an uncontrolled component, after passing the initial props we extract what we need with refs and let it go. If we need to update the header or tabbar accordingly to the index, we use the context, or some upper tree context (like in the `onIndexChange` example).

To support this kind of behavior (dynamic tabs or any kind of rerender), we need to redesign a little bit. Looks like you already figured everything out for the container. Now we need to rethink the tab bar.

andreialecu:
https://user-images.githubusercontent.com/697707/106937433-6bcdce00-6726-11eb-8056-04403a9d3674.mov

Still an issue with lazy tabs scroll sync. The non lazy variant doesn't seem to have it. Probably an easy fix.

PedroBern:
I'm working on the tabs.

PedroBern:
> Seems I also broke the Remove Tab functionality somehow in recent commits (from the dynamic example)

Only if `lazy={false}`

PedroBern:
I think we can leave the dynamic tabs support on the tab bar for a new PR, after merging this one. Everything seems to be working (besides the lazy tab sync).

I will leave some notes here:

#### TabBar (v4)

- [ ] can probably be a pure function of tabProps an indexDecimal
- [ ] support adding new tabs
- [ ] support removing tabs
- [ ] support rearranging the tabs
  - [ ] with same number of tabs
  - [ ] with more
  - [ ] with less
- [ ] ... everything that's already supported
  - [ ] scrollable tabs
  - [ ] non-scrollable tabs
  - [ ] scroll to current index on swipe if it's outside visible window

andreialecu:
The dynamic tabs look pretty solid now with the latest changes.

I can still reproduce this problem though: https://github.com/PedroBern/react-native-collapsible-tab-view/pull/93#issuecomment-773510738

Not exactly sure why. It's only in that "Lazy" example. In the dynamic example it works properly.

andreialecu:
The latest commit also shows a scroll sync issue. In the dynamic tabs example if you add a bunch of tabs, then scroll down so the header partially shows and then tap Shuffle tabs repeatedly, it will sometimes end up in a state that looks similar to this:

IMG_CFF058967996-1

andreialecu:
This seems to be called correctly so I'm not sure what the deal is:

https://github.com/andreialecu/react-native-collapsible-tab-view/blob/b0fbcb7c56fe9151d2861e2899683d63c2739127/src/createCollapsibleTabs.tsx#L881-L884

PedroBern:
I can reproduce both issues too

PedroBern:
> This seems to be called correctly so I'm not sure what the deal is:
>
> https://github.com/andreialecu/react-native-collapsible-tab-view/blob/b0fbcb7c56fe9151d2861e2899683d63c2739127/src/createCollapsibleTabs.tsx#L881-L884

Maybe the ref is being recreated on `useAnimatedDynamicRefs` and when the scroll sync happens, it's using the wrong ref.

danielfein:
Sorry if this is a bit of a lame question but... with these latest commits, how should I get the latest version? I'm going to try to render tabs conditionally with this proposed change and looks like everything was approved but not sure how that ends up as an updatable package! Thank you

andreialecu:
This is mostly complete, but a big breaking change. You'll need to take a look at the examples to see how to use it for now. It's actually way simpler than the current version.

To use it, if you use `yarn`, try changing your dependency to:
```
"react-native-collapsible-tab-view": "https://github.com/andreialecu/react-native-collapsible-tab-view.git#refactor-apicleanup3`
```

Then run `yarn install`. In theory it should build it download it from git and build it for you (via the `prepack` script)

andreialecu:
Some todos:

- [ ] update documentation script
- [ ] update documentation
- [ ] add example for flatlist/scrollview ref (the one in #99, but more polished)
- [ ] update changelog with the breaking changes
- [ ] consider whether to transform into a controlled component instead of controlling via the imperative ref handler
- [ ] make `useTabsContext` hook private and instead export other individual hooks as needed

List of breaking changes:

- `createCollapsibleTabs()` no longer exists. Instead the library exports different components (`Container`, `Tab`, `ScrollView`, `ListView`) which can be nested like normal react components.
- `refMap` no longer exists, and creating refs is no longer necessary (it's all handled internally)
- ScrollView/FlatList now forward their refs, so the caller can just tap into their `ref` prop
- the existing `onIndexChange` was renamed to `onTabChange`, and a separate `onIndexChange` was implemented that just returns the new index (as a `number`)
- `diffClampEnabled` was changed to `headerStickyness: 'disabled' | 'reveal-on-scroll'`

alexpchin:
@andreialecu @PedroBern Great work. How do we implement a `ScrollToTop` using the ref of the individual tabs?

andreialecu:
@alexpchin see the small demo in https://github.com/PedroBern/react-native-collapsible-tab-view/pull/99

You can just use `ref` directly from the ScrollView or FlatList.

danielfein:
'm not using in a production app so I don't mind if it's incomplete, happy to report bugs here. What's the best way to use these latest changes?

andreialecu:
@danielfein I already replied to your previous comment here with instructions. Scroll up a bit. Feedback would be greatly appreciated.

There are some issues on android with the latest commit which are being worked on (related to snapping and flatlists)

alexpchin:
> @danielfein I already replied to your previous comment here with instructions. Scroll up a bit. Feedback would be greatly appreciated.
> 
> There are some issues on android with the latest commit which are being worked on (related to snapping and flatlists)

Hi @andreialecu did as you said to use your version in my project. However, it seems like the `prepack` script doesn't automatically run. When adding this to my `postinstall` script.

```
npm i --prefix ./node_modules/react-native-collapsible-tab-view && npm run prepack --prefix ./node_modules/react-native-collapsible-tab-view
```

I see these errors:

```
> react-native-collapsible-tab-view@3.7.1 prepack /Users/alexchin/ReactNative/iynk-react-app/node_modules/react-native-collapsible-tab-view
> bob build

ℹ Building target commonjs
ℹ Cleaning up previous build at lib/commonjs
ℹ Compiling 15 files in src with babel
✓ Wrote files to lib/commonjs
ℹ Building target module
ℹ Cleaning up previous build at lib/module
ℹ Compiling 15 files in src with babel
✓ Wrote files to lib/module
ℹ Building target typescript
ℹ Cleaning up previous build at lib/typescript
ℹ Generating type definitions with tsc
src/helpers.tsx:18:14 - error TS2742: The inferred type of 'AnimatedFlatList' cannot be named without a reference to 'react-native-collapsible-tab-view/node_modules/@types/react'. This is likely not portable. A type annotation is necessary.

18 export const AnimatedFlatList = Animated.createAnimatedComponent(FlatList)
                ~~~~~~~~~~~~~~~~

src/hooks.tsx:34:17 - error TS2742: The inferred type of 'useContainerRef' cannot be named without a reference to 'react-native-collapsible-tab-view/node_modules/@types/react'. This is likely not portable. A type annotation is necessary.

34 export function useContainerRef() {
                   ~~~~~~~~~~~~~~~

src/hooks.tsx:481:17 - error TS2742: The inferred type of 'useSharedAnimatedRef' cannot be named without a reference to 'react-native-collapsible-tab-view/node_modules/@types/react'. This is likely not portable. A type annotation is necessary.

481 export function useSharedAnimatedRef<T extends RefComponent>(
                    ~~~~~~~~~~~~~~~~~~~~

src/index.tsx:27:14 - error TS2742: The inferred type of 'Tabs' cannot be named without a reference to 'react-native-collapsible-tab-view/node_modules/@types/react'. This is likely not portable. A type annotation is necessary.

27 export const Tabs = {
                ~~~~


Found 4 errors.

```

Also:

- [ ] This package has been renamed to 'react-native-builder-bob'. Please use it instead.

Propose releasing as @next ?

andreialecu:
[react-native-collapsible-tab-view-v3.7.1.tgz.zip](https://github.com/PedroBern/react-native-collapsible-tab-view/files/5952255/react-native-collapsible-tab-view-v3.7.1.tgz.zip)

I ran `pack` on the current tip of this PR. Extract the zip, and add it as `file:react-native-collapsible-tab-view-v3.7.1.tgz` (or check npm's documentation)

(or clone this PR, and run `yarn && yarn pack` to generate the file above)

andreialecu:
There is a known issue with RefreshControl + FlatList + Snapping on Android (iOS works fine). We believe it's a reanimated bug: https://github.com/software-mansion/react-native-reanimated/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc

alexpchin:
Hi @andreialecu I'm now setup with the new version in my project. Just getting to grips with the new setup but currently when I scroll up on iOS, the header does not scroll with the list. Also I'm seeing `[Object Map Iterator]` in the TabBar as per:

andreialecu:
Make sure to use the scrollview or flatlist exported from this package. Not the react native views one.

If the problem persists, a code sample of how you're using it would be helpful.

Issue #101, opened 2021-02-09
[Object Map Iterator] in Tabs


alexpchin:
Hi @PedroBern @andreialecu 

When upgrading to `4.0.0-next.0` and running the basic version in a react-native application (not expo) as per:

```ts
import { Tabs } from 'react-native-collapsible-tab-view';
.
.
.

return (
  <Tabs.Container HeaderComponent={_renderHeader}>
      <Tabs.Tab name="A">
        <Tabs.FlatList
          data={[0, 1, 2, 3, 4]}
          keyExtractor={(v) => v + ''}
          renderItem={null}
        />
      </Tabs.Tab>
      <Tabs.Tab name="B">
        <Tabs.ScrollView>{null}</Tabs.ScrollView>
      </Tabs.Tab>
 </Tabs.Container>
)
```

I see [Object Map Iterator], where the tabs are meant to be:

<imc src="https://user-images.githubusercontent.com/40461/107409517-2fbeb280-6b04-11eb-8d0e-8354d9bc75a2.png" width="300px" />

I also see:

<imc src="https://user-images.githubusercontent.com/40461/107409729-6399d800-6b04-11eb-98a1-d04215560793.png" width="300px" />

I'm also having a problem scrolling the header. I will add additional notes here as I investigate.

andreialecu:
I created a bare rn app here and cannot reproduce this:
https://github.com/andreialecu/rnctv-react-native-bare-test

![simulator_screenshot_D86B66F8-CBB1-42F3-AD25-EAAD071BC2D9](https://user-images.githubusercontent.com/697707/107497243-ff762300-6b9a-11eb-8ea4-93f815c456a5.png)

What might be different between the bare repro above and yours?

alexpchin:
I also do this in PR #101 and it also seemed to be ok for me, I'm trying to get to the bottom of the problem. It's strange because it happens on a screen in that just implements the QuickStart example. I will fork your repo and perhaps add the packages that I am using and update this issue if I find anything.

andreialecu:
Might be related to babel. Do you use any other babel (or metro) plugins?

andreialecu:
> I'm also having a problem scrolling the header. I will add additional notes here as I investigate.

@alexpchin I just realised on the bare rn app that using `2.0.0-rc.3` of `reanimated` breaks pretty much everything. Please try going back to `rc.0` and testing things with that version.

[...]

PR #107, opened 2021-02-10
feat: resolving object map iterator

alexpchin:
Seems to fix the issue #101

### DIFF
--- a/src/hooks.tsx
+++ b/src/hooks.tsx
@@ -79,15 +79,10 @@ export function useTabProps<T extends TabName>(
     })
     return tabOptions
   }, [children, tabType])
-  const optionEntries = [...options.entries()]
-  const optionKeys = [...options.keys()]
-
+  const optionEntries = Array.from(options.entries())
+  const optionKeys = Array.from(options.keys())
   const memoizedOptions = useDeepCompareMemo(() => options, [optionEntries])
-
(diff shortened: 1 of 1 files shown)


andreialecu:
Interesting. These should be equivalent. I wonder what the explanation is.

I was in the process of setting it up in a bare project to test it myself. Will come back to this in a bit.

alexpchin:
@andreialecu Yeah it's weird... I set it up as a bare app in #103 which actually seemed ok? But it's definitely showing as [Object Map Iterator] without?

andreialecu:
Let's continue this in #101.

I'm not opposed to this PR but it seems more of a refactoring than a fix. The refactoring makes sense.

alexpchin:
@andreialecu I'm still seeing this in `4.0.0-rc.2` when this change hasn't been made. I've cleared the Simulator and blasted all of the cache too.

andreialecu:
It's fine to merge. I still wonder what the problem is because spreads are used in other places so it may break things for you. Did you check for babel interactions?

andreialecu:
Merge, but you'll have to wait for @PedroBern to prepare a new release.

alexpchin:
> It's fine to merge. I still wonder what the problem is because spreads are used in other places so it may break things for you. Did you check for babel interactions?

I did. It's a really strange one. I want to know what is causing it. I will continue to look.

PR #121, opened 2021-02-11
feat: expose hooks

andreialecu:
Todo: add some examples

(currently untested)


andreialecu:
We could also add a hook for changing tabs, similar to [useNavigation](https://reactnavigation.org/docs/use-navigation/) from react-navigation. This would make it easier to navigate to other tabs from within tabs.

Issue #148, opened 2021-02-26
When `headerHeight` is not set sometimes tab opacity remains at 0


alexpchin:
### Current behavior

When using dynamic tabs and you don't add `headerHeight`, then sometimes (not every time) the Tabs will appear empty because the opacity will still be 0. Please see video:

[blank-tab.mov.zip](https://github.com/PedroBern/react-native-collapsible-tab-view/files/6046861/blank-tab.mov.zip)

If you add any value for headerHeight, i.e. `1` the content will visibly adjust but the content will be visible.

### Expected behaviour

You should not see the white screen.

andreialecu:
Tested this in the Example app and was not able to reproduce it. 

Please share a repro.

alexpchin:
Hi, @andreialecu  

I've been working to try to resolve this today... 
Basically, what I was seeing was that (only sometimes) the pager's opacity visibly remains at 0. 

(I can see the outline of the content in the siimulator inspector).

<img width="412" alt="Screenshot 2021-02-27 at 15 55 03" src="https://user-images.githubusercontent.com/40461/109392599-f0220400-7914-11eb-9a73-c6b920c2069d.png">

After doing some logging, I found that it happens when I see:

```
initialHeaderHeight --> undefined
initialTabName --> undefined
headerHeight --> undefined
pagerOpacity.value --> 0
update --> false
```

In the code below (348):

```js
// fade in the pager if the headerHeight is not defined
    useAnimatedReaction(
      () => {
        return (
          (initialHeaderHeight === undefined || initialTabName !== undefined) &&
          headerHeight !== undefined &&
          pagerOpacity.value === 0
        )
      },
      (update) => {
        if (update) {
          pagerOpacity.value = withTiming(1)
        }
      },
      [headerHeight]
    )
```

Followed shortly after by something like:

```
initialHeaderHeight --> undefined
initialTabName --> undefined
headerHeight --> undefined
pagerOpacity.value --> 0.00617283975268588
```

Which I think means there is a race condition between the logic there and in Container line `214`

```
pagerOpacity.value = withTiming(1)
```

Can 214 this be removed?

I "think" this will remove the race condition and fix the bug.

alexpchin:
Hi @andreialecu There seems to be a scenario where headerHeight is undefined but pager.opacity is not 0 so the pager stays at 0 opacity? Does withTiming(1) start at 0 and conclude at 1? Should this just be 1?

[...]

PR #149, opened 2021-02-27
feat: removing duplicate pagerOpacity.value change

alexpchin:
Fix for #148

### DIFF
--- a/src/Container.tsx
+++ b/src/Container.tsx
@@ -210,8 +210,6 @@ const Container = React.forwardRef<CollapsibleRef, CollapsibleProps>(
               false
             )
           })
-
-          pagerOpacity.value = withTiming(1)
         }
       },
       [tabNamesArray, refMap, afterRender, contentInset]
--- a/src/Container.tsx
+++ b/src/Container.tsx
@@ -173,7 +173,6 @@ const Container = React.forwardRef<CollapsibleRef, CollapsibleProps>(
 
     const afterRender = useSharedValue(0)
     React.useEffect(() => {
-      if (!firstRender.current) pagerOpacity.value = 0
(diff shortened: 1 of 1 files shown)


andreialecu:
It seems odd that this would make any difference, it's being set to `1`, which means `100%`. Even if there's a race condition, the result should still be that it's set to `100%` opacity.

🤔

andreialecu:
The change in this PR actually breaks the dynamic tabs example.

alexpchin:
If you also remove line 176.

```
if (!firstRender.current) pagerOpacity.value = 0
```

Then the example works.

Removing a tab gives `scrollToIndex out of range: requested index 1 but maximum is 0`. However, I think this was happening before these changes?

alexpchin:
I think it is ok?

andreialecu:
> If you also remove line 176.

That would be incorrect. The fade is on purpose otherwise the scroll position will jump around while the tabs are recreated. This is more visible on Android while the scrollview is slightly scrolled down so that only part of the header is visible.

`withTiming(1)` if the starting opacity is `1` shouldn't do anything. It's supposed to start from the existing value and transition to the new value.

I don't see how the fade can result in this problem, I suspect it's somewhere else. Without an actual repro I can see it's hard to verify though.

andreialecu:
Closing, feel free to ping us to reopen if this is still an issue.

alexpchin:
Hi @andreialecu No need to re-open this PR however, as per #148 I do still see that the page opacity sometimes doesn't turn to 1 when headerHeight isn't set.

My workaround at the moment is to set the `headerHeight` to 1, it is then re-calculated. This can cause some layout jumps but it does seem to stop the opacity problem.

PR #186, opened 2021-06-18
perf: optimize and reduce rerenders

andreialecu:
This should greatly improve performance, especially with scrollable tab bars.

It also optimizes various other things, and reduces rerenders a bit.

It should be possible to test it by adding a `git` dependency directly:

```
    "react-native-collapsible-tab-view": "PedroBern/react-native-collapsible-tab-view#perf-rerenders",
```


andreialecu:
This is now available as `react-native-collapsible-tab-view@4.2.1-rc.0` on npm for testing.

Seems to greatly improve performance especially on Android and lower end devices in a real world app I'm working on.

PR #205, opened 2021-10-12
fix: tabs should not scroll to top

### LINKED ISSUE #198: Tab switching & state updating causing scroll position issues
Hi there, hope you're all well. Love the library but I'm experiencing two problems at the moment, the latter of which is the most concerning. Would really appreciate the help!

### PROBLEM 1
When navigating between tabs, the scroll position of the inactive tab should persist however this only works if I don't interact with the active tab. When interacting with the active tab, the inactive tab's scroll position returns to 0. This may be expected behaviour, and if so, I'd really appreciate some suggestions!

https://user-images.githubusercontent.com/74291167/131118051-16a17ac9-6f15-4123-8cdb-566e2e481ad4.mp4

### PROBLEM 2
When there's a state change, the inactive tab's scroll position matches the current scroll position of the active tab. Very strange occurrence.

https://user-images.githubusercontent.com/74291167/131118066-3d424b3f-6295-4e45-ab49-80fc2531abae.mp4

### CODE

```javascript
import { Button, StyleSheet, View } from "react-native";
import React, { useState } from "react";

import { Tabs } from "react-native-collapsible-tab-view";

const CollapsibleTabs = () => {
  const [count, setCount] = useState(0);

  return (
    <View style={styles.container}>
      <View style={styles.button}>
        <Button
          title={`Update State: ${count}`}
          onPress={() => setCount(count + 1)}
        />
      </View>
      <Tabs.Container renderHeader={Header} >
        <Tabs.Tab name="One">
          <Tabs.ScrollView>
            <Page />
          </Tabs.ScrollView>
        </Tabs.Tab>
        <Tabs.Tab name="Two">
          <Tabs.ScrollView>
            <Page />
          </Tabs.ScrollView>
        </Tabs.Tab>
      </Tabs.Container>
    </View>
  );
};

const Header = (props) => {
  return <View {...props} style={styles.header} />
}

const Page = () => {
  return (
    <View style={styles.page}>
      <View style={{ width: 50, height: 50, backgroundColor: "green" }} />
      <View style={{ width: 50, height: 50, backgroundColor: "blue" }} />
      <View style={{ width: 50, height: 50, backgroundColor: "purple" }} />
      <View style={{ width: 50, height: 50, backgroundColor: "violet" }} />
      <View style={{ width: 50, height: 50, backgroundColor: "pink" }} />
      <View style={{ width: 50, height: 50, backgroundColor: "red" }} />
      <View style={{ width: 50, height: 50, backgroundColor: "orange" }} />
      <View style={{ width: 50, height: 50, backgroundColor: "yellow" }} />
    </View>
  );
};

const styles = StyleSheet.create({
  container: {
    flex: 1,
  },
  button: {
    position: "absolute",
    bottom: 30,
    right: 30,
    zIndex: 1
  },
  background: {
    position: "absolute",
    top: 0,
    width: 390,
    height: 390,
  },
  page: {
    width: 390,
    height: 830 * 2,
    backgroundColor: "black",
  },
  header: {
    width: 390,
    height: 390,
    backgroundColor: "pink",
  },
});

export default CollapsibleTabs;

```

andreialecu:
Fixes #198

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
   "peerDependencies": {
     "react": "*",
     "react-native": "*",
-    "react-native-reanimated": "~2.0.0"
+    "react-native-reanimated": ">=2.1.0"
   },
   "husky": {
     "hooks": {
--- a/src/Container.tsx
+++ b/src/Container.tsx
@@ -235,9 +235,7 @@ export const Container = React.memo(
       // the next index in advance
       useAnimatedReaction(
         () => {
-          const nextIndex = isSwiping.value
-            ? Math.round(indexDecimal.value)
-            : null
+          const nextIndex = Math.round(indexDecimal.value)
           return nextIndex
         },
         (nextIndex) => {
@@ -386,7 +384,7 @@ export const Container = React.memo(
           // when is scrolling or gliding.
           if (!isScrolling.value && !isGliding.value) {
             const i = tabNames.value.findIndex((n) => n === name)
-            calculateNextOffset.value = i
+
             if (name === focusedTab.value) {
               const ref = refMap[name]
               runOnUI(scrollToImpl)(
@@ -516,7 +514,6 @@ export const Container = React.memo(
             </Animated.View>
             {headerHeight !== undefined && (
               <AnimatedFlatList
-                // @ts-expect-error problem with reanimated types, they're missing `ref`
                 ref={containerRef}
                 initialScrollIndex={index.value}
                 data={data}
--- a/src/FlatList.tsx
+++ b/src/FlatList.tsx
@@ -20,13 +20,7 @@ import {
 const FlatListMemo = React.memo(
   React.forwardRef<RNFlatList, React.PropsWithChildren<FlatListProps<unknown>>>(
     (props, passRef) => {
-      return (
-        <AnimatedFlatList
-          // @ts-expect-error problem with reanimated types, they're missing `ref`
-          ref={passRef}
-          {...props}
-        />
-      )
+      return <AnimatedFlatList ref={passRef} {...props} />
     }
   )
 )
--- a/src/SectionList.tsx
+++ b/src/SectionList.tsx
@@ -22,13 +22,7 @@ const SectionListMemo = React.memo(
     RNSectionList,
     React.PropsWithChildren<SectionListProps<unknown>>
   >((props, passRef) => {
-    return (
-      <AnimatedSectionList
-        // @ts-expect-error problem with reanimated types, they're missing `ref`
-        ref={passRef}
-        {...props}
-      />
-    )
+    return <AnimatedSectionList ref={passRef} {...props} />
   })
 )
 
--- a/src/hooks.tsx
+++ b/src/hooks.tsx
@@ -203,7 +203,9 @@ export function useScroller<T extends RefComponent>() {
     ) => {
       'worklet'
       if (!ref) return
-      // console.log(`${_debugKey}, y: ${y}, y adjusted: ${y - contentInset}`)
+      // console.log(
+      //   `${_debugKey}, y: ${y}, y adjusted: ${y - contentInset.value}`
+      // )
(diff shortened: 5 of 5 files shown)


PedroBern:
@andreialecu I got a bug at first glance on old devices - O don't have this bug on main.


https://user-images.githubusercontent.com/39778068/137044422-48be1c0a-d45d-4bb7-aacf-8a67aaf93429.mp4

Hirbod:
Anything holding back to merge this?

andreialecu:
I've been using this for a while in production and it seems to work fine. I'll merge it.

andreialecu:
Released as 4.5.0

Hirbod:
@andreialecu weird, for me, it is still resetting the scroll position, as soon as I switch the tab.

andreialecu:
Indeed there's an edge case here, it works if the tab you're moving away from is scrolled down a bit, but if one is scrolled to the top, then it will reset the others.

I'll open a subsequent PR to fix this.

Hirbod:
@andreialecu here is a video demonstrating, that this haven't changed anything:
https://streamable.com/wr7mrh

As soon as I change the tab, my previous view will scroll to top, which is pretty annoying, because I would love to maintain the scroll position (which is also pretty important for FlatList performance, as it currently retriggers all the VirzualizedList events)

Hirbod:
For me, it only works, when every lazy loaded tab was loaded and I scrolled within every single one of them once.

andreialecu:
@Hirbod https://github.com/PedroBern/react-native-collapsible-tab-view/pull/230 should fix it, it was just released as 4.5.2

Hirbod:
@andreialecu it's better but still not 100% fixed.

I have 4 tabs (with prop lazy). As soon as the lazy one loads, it syncs all the other screens.
See following video (i have added indexes to the views):

So there is still an edge case to be handled I guess.
https://streamable.com/h184la

andreialecu:
I'm not able to reproduce this. See this recording from the example app:

https://user-images.githubusercontent.com/697707/149623449-172d9b8a-783b-4de5-bc2c-3b6c0f73aff4.mov

Hirbod:
Sorry, here is the updated video. 
https://streamable.com/vk9n8t

andreialecu:
Oh, ok, I see now.

Hirbod:
@andreialecu And there is another edge case. It will restore all scroll positions to top as soon as I fully scroll up (uncollapsing the header)
https://streamable.com/b1bug4

Would be wonderful if you could fix those last two remaining little bugs, then this lib is perfect :D

Hirbod:
@andreialecu did you had a chance to have a last look at the issue? Putting $50 bounty on it!

hirbod:
@andreialecu any chance?

andreialecu:
> @andreialecu And there is another edge case. It will restore all scroll positions to top as soon as I fully scroll up (uncollapsing the header) [streamable.com/b1bug4](https://streamable.com/b1bug4)
> 
> Would be wonderful if you could fix those last two remaining little bugs, then this lib is perfect :D

If one tab is at the top, with the header revealed, then if you switch to a different tab then the header is still revealed, and it only makes sense to be scrolled to the top in that tab as well.

 If that didn't happen, the flatlist would look glitched. I'm not entirely sure what you'd expect to happen here, care to elaborate?

hirbod:
@andreialecu I agree with the last edge case, more likely I am still stumbling across this bug:



> Sorry, here is the updated video. https://streamable.com/vk9n8t
> 
> You can cleary see, the second tab 3 got rendered, tab 1 and tab 2 had the same indexes at top (like the third one coming in). Same happens when the 4th tab renderes. After that, the sync works fine.

andreialecu:
Are you able to reproduce this in the example app that comes with this repository? If you can, can you share a video that shows it on the example app?

hirbod:
I'll need to set it up. Will get back to you after its running on my device

PR #211, opened 2021-11-14
Feature/stretchable

filipengberg:


andreialecu:
Interestingly I think I've just implemented this in #228

filipengberg:
Thats great!

I updated this PR to only contain the stretchable header example using the prop you added.

The reason I never got to finishing my PR was I encountered an issue where the header would not stretch if you drag down when the list is at its top position. You have to first drag to scroll the list a little bit downwards and then up and the overscroll works. Its seems this issue is still present with your implementation (I see in the example video in #228 that it starts by scrolling down a little bit before scrolling up). Do you have any idea what the reason for this could be and how to fix it? I'd be happy to try to help out fixing it (spent an hour last time but did not manage to find any solution)

andreialecu:
I think that may only happen with Expo for some reason. In my real app I can scroll down with no issues so I didn't investigate it too much.

Do you have that issue in non expo apps?

filipengberg:
@andreialecu You're absolutely right, it does work when I test it in another app. Thats weird, but thanks a lot for pointing that out! Feel free to merge this PR if you want the added example, otherwise you can just close it.

PR #230, opened 2022-01-15
fix: edge case with tabs still scrolling to top

andreialecu:
ref #205

### DIFF
--- a/src/hooks.tsx
+++ b/src/hooks.tsx
@@ -7,6 +7,7 @@ import {
   MutableRefObject,
   useEffect,
 } from 'react'
+import { StyleSheet } from 'react-native'
 import { ContainerRef, RefComponent } from 'react-native-collapsible-tab-view'
 import Animated, {
   cancelAnimation,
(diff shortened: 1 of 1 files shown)


andreialecu:
Released as 4.5.2

Hirbod:
@andreialecu it's better but still not 100% fixed.

I have 4 tabs (with prop lazy). As soon as the lazy one loads, it syncs all the other screens.
See following video (i have added indexes to the views):

So there is still an edge case to be handled I guess.
https://streamable.com/h184la

andreialecu:
Hey @hirbod, I've been refactoring some things over in #255 and I think this may be fixed. Could you give `react-native-collapsible-tab-view@5.0.0-rc.3` a try?

See the PR above for prerequisites.

hirbod:
Will be able to test this in about 2 hours. Does it also handle the issue that happens on android with numColums on FlatList?

andreialecu:
Not sure, I don't use `numColumns` but it might.

hirbod:
@andreialecu I can tell you that I've upgraded and the first thing I realized is that that new release is breaking the swipe behavior in my view. I can only swipe tabs now when I place my finger on the tabs directly or when I swipe in the header area. 

It does not play nicely with `RectButton` by react-native-gesture-handler anymore. It does work with TouchableOpacity though

PR #249, opened 2022-03-27
Update TabReactElement - allow null as Tab

hirbod:
Without this, TypeScript linter won't allow conditional tabs with a `null` return.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -56,7 +56,7 @@ export type OnTabChangeCallback<T extends TabName = TabName> = (
 ) => void
 
 export type TabReactElement<T extends TabName = TabName> = React.ReactElement<
-  TabProps<T>
+  TabProps<T> | null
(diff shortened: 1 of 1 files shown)


martymfly:
> Without this, TypeScript linter won't allow conditional tabs with a `null` return.

This didn't work for me. I still get this ts error. Any other workaround?

Type 'Element | null' is not assignable to type 'TabReactElement<TabName>'.
  Type 'null' is not assignable to type 'ReactElement<TabProps<TabName>, string | JSXElementConstructor<any>>'

andreialecu:
I pulled in this change into the latest RC in https://github.com/PedroBern/react-native-collapsible-tab-view/pull/255.

Also fixed the other TypeScript error.

It's already on npm as `5.0.0-rc.9` if you'd like to try it

PR #285, opened 2022-07-26
fix:flatlist list-item not pressed when keyboard is open

026binit:
- Added `keyboardShouldPersistTaps="handled"` inside `Flatlist.js`


andreialecu:
You should be able to do `<Tabs.FlatList keyboardShouldPersistTaps="handled" ... />` instead.

PR #319, opened 2023-02-25
Fix compatibility issues with Reanimated v3

frw:
# What's Changed
Reanimated v3 has brought some breaking changes. Fortunately, the changes that affect this library are not too significant.

- `Animated.Extrapolate` is no longer available. Use `Extrapolate` import directly from `react-native-reanimated`.
- All functions passed into `runOnUI` has to be marked as a `'worklet'`.


andreialecu:
Thanks for this. Found some things missing while looking into it and I have pushed #322 with some additional updates.

Released reanimated 3 support as 6.0.0-rc.0

PR #333, opened 2023-04-21
fix: FlashList re-render issue

lucianobracco-geojam:


andreialecu:
The fix would reintroduce the issue with the message being logged in all apps.

I opened a PR with an improved fix here:
https://github.com/PedroBern/react-native-collapsible-tab-view/pull/334

PR #351, opened 2023-07-14
Adding MasonryFlashList

Jonatthu:
Adds support to MasonryFlashList from flash list.


Jonatthu:
https://github.com/PedroBern/react-native-collapsible-tab-view/pull/353

Issue #335, opened 2023-04-22
Flashlist issue with minHeight


andreialecu:
Flashlist does not support `minHeight` in `contentContainerStyle` resulting in an issue if there isn't enough content to fill the screen when swiping between panes.

See video below for an explanation:

https://user-images.githubusercontent.com/697707/233771319-5087ff0b-edd3-40a2-9d14-98c6cbb4f1b4.mp4

andreialecu:
Can be repro'd by uncommenting these lines in the example:

https://github.com/PedroBern/react-native-collapsible-tab-view/blob/834dc4b9c0edc93feaa8ff4affa93f266a248232/example/src/Shared/ExampleComponentFlashList.tsx#L33-L34

gkartalis:
Not sure if helpful but:
- I think flashlist needs a height all the time in order to correctly render the list.
- if you wrap the `react-native-collapsible-tab-view/example/src/Shared/ContactsFlashList.tsx` with a View with fixed height and subtract the `const { top } = useHeaderMeasurements()`  like this:
```ts
<View style={{ height: height - top.value }} >
<Tabs.FlashList>
...
</View>
```

where height is const `height = useWindowDimensions().height` at least it is stuck on the previous tab scroll value. 🤷‍♂️ hope it helps

https://github.com/PedroBern/react-native-collapsible-tab-view/assets/21178754/242adc6d-af2a-433f-9d6c-8150757e5132

ngocan3897:
i got this issue hope will have solution soon

zRelux:
Any update?

zRelux:
You can bypass this by wrapping the flashlist in the tabs scrollview

senghuotlay:
I tried, it doesn't seems to work at all

beh0907:
> You can bypass this by wrapping the flashlist in the tabs scrollview

But 'onRefresh' and 'onEndReached' are not working

senghuotlay:
I am very tempted to use this https://github.com/netguru/sticky-parallax-header

omotoshogyro:
> Not sure if helpful but:
> 
> * I think flashlist needs a height all the time in order to correctly render the list.
> * if you wrap the `react-native-collapsible-tab-view/example/src/Shared/ContactsFlashList.tsx` with a View with fixed height and subtract the `const { top } = useHeaderMeasurements()`  like this:
> 
> ```ts
> <View style={{ height: height - top.value }} >
[...]

PR #353, opened 2023-07-17
feat: masonry flashlist support

gkartalis:
# Description

Adds masonry flashlist support to the library.

Bonus: updates `@shopify/flash-list` to latest

Unfortunately though the issue https://github.com/PedroBern/react-native-collapsible-tab-view/issues/335 is the same with MasonryFlashList.

Example demo Video: 

https://github.com/PedroBern/react-native-collapsible-tab-view/assets/21178754/a0c4c8f3-ac3a-4cfb-a555-e0493861dd79


example/ios/Podfile.lock
```diff
@@ -411,7 +411,7 @@ DEPENDENCIES:
   - React-RCTVibration (from `../node_modules/react-native/Libraries/Vibration`)
   - React-runtimeexecutor (from `../node_modules/react-native/ReactCommon/runtimeexecutor`)
   - ReactCommon/turbomodule/core (from `../node_modules/react-native/ReactCommon`)
-  - "RNFlashList (from `../../node_modules/@shopify/flash-list`)"
```
gkartalis, line 414: question: was there a reason that we had `../../node_modules/@shopify/flash-list` previously? Is it to prevent the libary `react-native-collapsible-tab-view` from crashing in case that people do not have the dep installed or something?

andreialecu: Not sure.

This is just the example app so it won't affect the library when published, so it should be fine.

gkartalis: made it work I think, pushing changes in a bit

gkartalis: will also add some videos

andreialecu:
Thanks! Released as 6.2.0

PR #360, opened 2023-08-04
fix: recycling and re-rendering issues with Masonry

gkartalis:
## Description

When adding the lines for calling loadMore onEndReached of MasonryFlashlist we get this weird behavior in the example where the list flashes and returns to the top.

This PR fixes that but it requires exporting `recyclerlistviewUnsafe` from [MasonryFlashListRef](https://github.com/Shopify/flash-list/blob/main/src/MasonryFlashList.tsx#L79-L83).

This is happening due to us relying on the recyclerlistviewUnsafe while it not existsing on the MasonryFlashListRef interface. 

Opened a PR in flash-list about that in order to solve the re-rendering issues. ~~and we need to merge this PR after the related [flashlist pr](https://github.com/Shopify/flash-list/pull/892) gets merged and we bump the version here.~~ turns out we don't need the other PR! Realised it after cleaning up the patch! 

⚠️ ~~I have patched the shopify library on this PR just to make sure that this solution works and will clean everything up before merging this.~~

### Follow ups:

- [x] remove patch-package and clean patch
- [x] bump flashlist version for example
- [x] ~~merge flashlist [related pr](https://github.com/Shopify/flash-list/pull/892) before merging this one~~ - actually I think we can merge this as it is since it has the `@ts-expect-error` above the actual `recyclerlistview_unsafe` line that **protects it from crashing** but at the same time might not work without patching the library / using a fork to make it actually work if shopify doesn't review / reject the PR. 

|Before|After|
|---|---|
|<video src="https://github.com/PedroBern/react-native-collapsible-tab-view/assets/21178754/cf75b145-7ff7-4723-a1aa-35a8dfdb86f5" width="400" />|<video src="https://github.com/PedroBern/react-native-collapsible-tab-view/assets/21178754/75d50346-340f-48c8-b5f7-e959d9a5f9c2" width="400" />|


src/MasonryFlashList.tsx
```diff
@@ -117,20 +129,24 @@ function MasonryFlashListImpl<R>(
     [_contentContainerStyle, contentContainerStyle.paddingTop]
   )
 
+  const refWorkaround = useCallback(
+    (value: MasonryFlashListMemoRef | null): void => {
+      // https://github.com/Shopify/flash-list/blob/2d31530ed447a314ec5429754c7ce88dad8fd087/src/FlashList.tsx#L829
+      // We are not accessing the right element or view of the Flashlist (recyclerlistview). So we need to give
+      // this ref the access to it
+      ;(recyclerRef as any)(value?.recyclerlistview_unsafe)
```
gkartalis, line 137: this was not working properly before the patch and exporting it from the interface fixes the rendering issue! 🔥

gkartalis: this and the recyclerRef + useEffect recyclerRef additions

gkartalis:
@andreialecu let me know if you have any suggestions on how to proceed! Would love to see this merged because it solves many issues for us at Artsy! 

Turns out we don't explicitly need the shopify PR to get merged for this to work so they can be merged independently! 🎉

andreialecu:
Alright. Going to merge this now.

andreialecu:
Released as 6.2.1

PR #395, opened 2024-04-12
fix: indicator display for single tab

Kerumen:
In my app I added some container style with padding to have a specific display for the tabs. When there is only one tab, the indicator is not shifted accordingly. This is due to the `itemsLayout.length > 1` condition and the `undefined` value of `transform`. The same way as it's done for the `width`, I changed the `transform` to match the first item of the `itemsLayout` array if there is only one tab.

<table>
<tr>
 <th> Before
 <th> After
<tr>
 <td> 
 <td> 
</table>

### DIFF
--- a/src/MaterialTabBar/Indicator.tsx
+++ b/src/MaterialTabBar/Indicator.tsx
@@ -19,19 +19,21 @@ const Indicator: React.FC<IndicatorProps> = ({
   const opacity = useSharedValue(fadeIn ? 0 : 1)
 
   const stylez = useAnimatedStyle(() => {
-    const transform =
-      itemsLayout.length > 1
-        ? [
-            {
-              translateX: interpolate(
+    const transform = [
+      {
+        translateX:
+          itemsLayout.length > 1
+            ? interpolate(
                 indexDecimal.value,
                 itemsLayout.map((_, i) => i),
                 // when in RTL mode, the X value should be inverted
                 itemsLayout.map((v) => (isRTL ? -1 * v.x : v.x))
-              ),
-            },
-          ]
-        : undefined
+              )
+            : isRTL
+              ? -1 * itemsLayout[0]?.x
+              : itemsLayout[0]?.x,
+      },
+    ]
 
     const width =
       itemsLayout.length > 1
--- a/src/MaterialTabBar/Indicator.tsx
+++ b/src/MaterialTabBar/Indicator.tsx
@@ -19,6 +19,8 @@ const Indicator: React.FC<IndicatorProps> = ({
   const opacity = useSharedValue(fadeIn ? 0 : 1)
 
   const stylez = useAnimatedStyle(() => {
+    const firstItemX = itemsLayout[0]?.x ?? 0
+
     const transform = [
       {
         translateX:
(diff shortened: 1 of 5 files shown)


src/MaterialTabBar/Indicator.tsx
```diff
@@ -19,19 +19,21 @@ const Indicator: React.FC<IndicatorProps> = ({
...
                 indexDecimal.value,
                 itemsLayout.map((_, i) => i),
                 // when in RTL mode, the X value should be inverted
                 itemsLayout.map((v) => (isRTL ? -1 * v.x : v.x))
-              ),
-            },
-          ]
-        : undefined
+              )
+            : isRTL
```
adrianurs, line 32:
```suggestion
            : isRTL && itemsLayout[0]?.x
```

Because on the line below you are doing -1 * itemsLayout[0]?.x, which can result nan so check also if it exists

src/MaterialTabBar/Indicator.tsx
```diff
@@ -19,19 +19,21 @@ const Indicator: React.FC<IndicatorProps> = ({
...
+              ? -1 * itemsLayout[0]?.x
```
adrianurs, line 33:
```suggestion
              ? -itemsLayout[0]?.x
```

adrianurs (commented):
I would also suggest to move the check from line 32 to line 34 outside from the current place because nested ternary operators might result confusing.
You can declare a variable before the transform object then use it.

`
const firstItemLayoutPosition = isRTL && itemsLayout[0]?.x ? -itemsLayout[0]?.x : itemsLayout[0]?.x
`

src/MaterialTabBar/Indicator.tsx
andreialecu: This suggestion doesn't seem correct. What if `x` is 0?

May need to check for `typeof itemsLayout[0]?.x === 'number'` instead

src/MaterialTabBar/Indicator.tsx
andreialecu: This is fine as originally written, other code in this file uses the `-1 *` pattern and it's more consistent that way.

andreialecu:
@Kerumen how did you add the padding style to see this issue?

I'm getting the following result without the need for this PR:

simulator_screenshot_7B1EBD32-E6CE-4AE3-B8BD-F288234A4E48

With this setup:

```tsx
<Tabs.Container
      renderHeader={Header}
      headerHeight={HEADER_HEIGHT}
      renderTabBar={(props) => (
        <MaterialTabBar
          {...props}
          scrollEnabled
          style={{ paddingHorizontal: 30 }}
        />
      )}
    >
      <Tabs.Tab name="article" label="Article">
        <Tabs.ScrollView>
          <ArticleContent />
        </Tabs.ScrollView>
      </Tabs.Tab>
    </Tabs.Container>
```

Kerumen:
If you have a lot of tabs, the last tab is cropped one when you put the `paddingHorizontal` as `style` on the `MaterialTabBar`. To fix it, I've used `contentContainerStyle` and it works fine but then there is the indicator issue when there is only one tab.

Screenshot 2024-04-15 at 10 49 08

andreialecu:
Merged. Thank you for the PR!

PR #423, opened 2024-06-18
refactor(container,lazy,hooks): avoid blocking JS thread by refactoring SharedValue.value access

AndreiCalazans:
Hey  @andreialecu  and @gkartalis (tagging bc I can't add reviewers), I was introducing this library to solve a UI pattern and noticed blocking JavaScript thread behavior that caused a regression of over 500ms to our initial screen render. 

cpu-profile-of-collapsible-tab-view

The flame graph above shows that the JavaScript thread is blocked by executeOnUIRuntimeSync for quite some time impacting the initial render time of a screen by over 500ms.

On closer analysis I found that the problem is caused by reading a SharedValue.value in the main JS thread. This library does this quite often in places that could avoid it. This PR refactors most of the places where it does not need to be a SharedValue while trying to not cause any regressions.

I understand there isn't unit or e2e tests for this library so I tried my best to test the Example app. I posted multiple recordings below of how it currently works with these changes.

[I also wrote a detailed blogpost explaining this Reanimated blocking behavior](https://andrei-calazans.com/posts/reanimated-blocking-js-thread/).

You can also inspect the before and after CPU profilings, [I stored them here](https://gist.github.com/AndreiCalazans/3fc4a522f94ea3387060f9c6fe461840/revisions). The CPU profile will show that before these changes there were 57 calls to executeOnUIRuntimeSync when navigating from the main menu to the FlashList example in the Example app and now there are only 14 calls.

| Before (57 calls) | After (14 calls) |
|--|--|
| image | image |

## Demos

| One | Two |
|--|--|
| <video src="https://github.com/PedroBern/react-native-collapsible-tab-view/assets/20777666/1f688b19-a427-428d-8fb8-0141d7f862ab"/> | <video src="https://github.com/PedroBern/react-native-collapsible-tab-view/assets/20777666/6b1b8e30-a7b9-4459-b83a-2175df02b4cc"/>  |

| Three | Four |
|--|--|
| <video src="https://github.com/PedroBern/react-native-collapsible-tab-view/assets/20777666/fd42dc3a-ef9b-4647-8d0e-3ab3eea26a24"/> | <video src="https://github.com/PedroBern/react-native-collapsible-tab-view/assets/20777666/40b58e5e-20a4-49e7-b357-0675da0d0cff"/>  |

| Five | Six |
|--|--|
| <video src="https://github.com/PedroBern/react-native-collapsible-tab-view/assets/20777666/ac4e036f-a7f0-47bc-b824-ce7411b69bb5"/> | <video src="https://github.com/PedroBern/react-native-collapsible-tab-view/assets/20777666/779df616-5b09-4d18-8f1a-79bcb6ae5eba"/>  |

| Seven | Eight |
|--|--|
| <video src="https://github.com/PedroBern/react-native-collapsible-tab-view/assets/20777666/130b7eaf-6245-4cdb-9451-e6e8f4256889"/> | <video src="https://github.com/PedroBern/react-native-collapsible-tab-view/assets/20777666/8615933a-4e99-4026-8b1f-5a9338150602"/>  |

### DIFF
--- a/src/Container.tsx
+++ b/src/Container.tsx
@@ -1,10 +1,5 @@
 import React from 'react'
-import {
-  LayoutChangeEvent,
-  StyleSheet,
-  useWindowDimensions,
-  View,
-} from 'react-native'
+import { StyleSheet, useWindowDimensions, View } from 'react-native'
 import PagerView from 'react-native-pager-view'
 import Animated, {
   runOnJS,
@@ -15,6 +10,7 @@ import Animated, {
   useSharedValue,
   withDelay,
   withTiming,
+  useFrameCallback,
 } from 'react-native-reanimated'
 
 import { Context, TabNameContext } from './Context'
@@ -27,6 +23,7 @@ import {
   useContainerRef,
   usePageScrollHandler,
   useTabProps,
+  useLayoutHeight,
 } from './hooks'
 import {
   CollapsibleProps,
@@ -93,25 +90,29 @@ export const Container = React.memo(
       const windowWidth = useWindowDimensions().width
       const width = customWidth ?? windowWidth
 
-      const containerHeight = useSharedValue<number | undefined>(undefined)
+      const [containerHeight, getContainerLayoutHeight] = useLayoutHeight()
 
-      const tabBarHeight = useSharedValue<number | undefined>(
-        initialTabBarHeight
-      )
+      const [tabBarHeight, getTabBarHeight] =
+        useLayoutHeight(initialTabBarHeight)
 
-      const headerHeight = useSharedValue<number | undefined>(
+      const [headerHeight, getHeaderHeight] = useLayoutHeight(
         !renderHeader ? 0 : initialHeaderHeight
       )
+      const initialIndex = React.useMemo(
+        () =>
+          initialTabName
+            ? tabNamesArray.findIndex((n) => n === initialTabName)
+            : 0,
+        [initialTabName, tabNamesArray]
+      )
 
-      const contentInset = useDerivedValue(() => {
+      const contentInset = React.useMemo(() => {
         if (allowHeaderOverscroll) return 0
 
         // necessary for the refresh control on iOS to be positioned underneath the header
         // this also adjusts the scroll bars to clamp underneath the header area
-        return IS_IOS
-          ? (headerHeight.value || 0) + (tabBarHeight.value || 0)
-          : 0
-      })
+        return IS_IOS ? (headerHeight || 0) + (tabBarHeight || 0) : 0
+      }, [headerHeight, tabBarHeight, allowHeaderOverscroll])
 
       const snappingTo: ContextType['snappingTo'] = useSharedValue(0)
       const offset: ContextType['offset'] = useSharedValue(0)
@@ -131,22 +132,16 @@ export const Container = React.memo(
         () => tabNamesArray,
         [tabNamesArray]
       )
-      const index: ContextType['index'] = useSharedValue(
-        initialTabName
-          ? tabNames.value.findIndex((n) => n === initialTabName)
-          : 0
-      )
+      const index: ContextType['index'] = useSharedValue(initialIndex)
 
       const focusedTab: ContextType['focusedTab'] =
         useDerivedValue<TabName>(() => {
           return tabNames.value[index.value]
         }, [tabNames])
-      const calculateNextOffset = useSharedValue(index.value)
+      const calculateNextOffset = useSharedValue(initialIndex)
       const headerScrollDistance: ContextType['headerScrollDistance'] =
         useDerivedValue(() => {
-          return headerHeight.value !== undefined
-            ? headerHeight.value - minHeaderHeight
-            : 0
+          return headerHeight !== undefined ? headerHeight - minHeaderHeight : 0
         }, [headerHeight, minHeaderHeight])
 
       const indexDecimal: ContextType['indexDecimal'] = useSharedValue(
@@ -167,7 +162,7 @@ export const Container = React.memo(
           scrollToImpl(
             refMap[name],
             0,
-            scrollYCurrent.value - contentInset.value,
+            scrollYCurrent.value - contentInset,
             false
           )
         }
@@ -213,6 +208,33 @@ export const Container = React.memo(
         [onIndexChange, onTabChange]
       )
 
+      const syncCurrentTabScrollPosition = () => {
+        'worklet'
+
+        const name = tabNamesArray[index.value]
+        scrollToImpl(
+          refMap[name],
+          0,
+          scrollYCurrent.value - contentInset,
+          false
+        )
+      }
+
+      /*
+       * We run syncCurrentTabScrollPosition in every frame after the index
+       * changes for about 1500ms because the Lists can be late to accept the
+       * scrollTo event we send. This fixes the issue of the scroll position
+       * jumping when the user changes tab.
+       * */
+      const toggleSyncScrollFrame = (toggle: boolean) =>
+        syncScrollFrame.setActive(toggle)
+      const syncScrollFrame = useFrameCallback(({ timeSinceFirstFrame }) => {
+        syncCurrentTabScrollPosition()
+        if (timeSinceFirstFrame > 1500) {
+          runOnJS(toggleSyncScrollFrame)(false)
+        }
+      }, false)
+
       useAnimatedReaction(
         () => {
           return calculateNextOffset.value
@@ -236,13 +258,14 @@ export const Container = React.memo(
               scrollYCurrent.value =
                 scrollY.value[tabNames.value[index.value]] || 0
             }
+            runOnJS(toggleSyncScrollFrame)(true)
           }
         },
         []
       )
 
       useAnimatedReaction(
-        () => headerHeight.value,
+        () => headerHeight,
         (_current, prev) => {
           if (prev === undefined) {
             // sync scroll if we started with undefined header height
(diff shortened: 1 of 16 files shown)


src/Container.tsx
```diff
@@ -313,11 +281,14 @@ export const Container = React.memo(
...
-          onTabPress(tabNamesArray[tabNamesArray.length - 1])
+      useAnimatedReaction(
+        () => tabNamesArray.length,
+        (tabLength) => {
+          if (index.value >= tabLength) {
+            runOnJS(onTabPress)(tabNamesArray[tabLength - 1])
+          }
         }
-      }, [index.value, onTabPress, tabNamesArray])
+      )
```
AndreiCalazans, line 291: This is essentially the same as was just instead of running inside a useEffect it runs inside a useAnimatedReaction to avoid blocking the JS thread.

src/Container.tsx
```diff
@@ -131,22 +131,16 @@ export const Container = React.memo(
...
-          : 0
-      )
+      const index: ContextType['index'] = useSharedValue(initialIndex)
 
       const focusedTab: ContextType['focusedTab'] =
         useDerivedValue<TabName>(() => {
           return tabNames.value[index.value]
         }, [tabNames])
-      const calculateNextOffset = useSharedValue(index.value)
+      const calculateNextOffset = useSharedValue(initialIndex)
```
AndreiCalazans, line 140: This change is safe to do because the initial value passed to useSharedValue will only be captured once on first pass, this is not reactive.

andreialecu:
Will look over this tomorrow. 

Thanks for contributing!

andreialecu:
Very interesting, I didn't know about this behavior of Reanimated. Is it documented anywhere?

Could it be a bug in Reanimated, maybe a recent regression? I don't remember running into an issue before or hearing about SharedValue blocking JS threads.

@piaskowyk could you kindly shed some light? See Andrei Calazan's blog post and reddit posts:
https://andrei-calazans.com/posts/reanimated-blocking-js-thread/
https://www.reddit.com/r/reactnative/comments/1diy3pc/reanimated_can_block_the_main_javascript_thread/

I remember introducing `SharedValue` here specifically to improve performance in the past. For example to be able to do this on the UI thread:
https://github.com/PedroBern/react-native-collapsible-tab-view/compare/v5.0.0-rc.1...v5.0.0-rc.10#diff-739a99e40f851900557fbc98026c870d8ac3c2da65ec6fc29d60c161effaf20dR240-R248

AndreiCalazans:
Hey @andreialecu, thanks for reviewing. 

I also didn't know this until now, not sure if it's a regression since the behavior kinda of makes sense if you think about it. Any time you need to read a SharedValue.value from within the JavaScript thread it needs to bridge the value synchronously from the HostObject so that can have a non-trivial cost (anywhere from 5 to 20ms depending on device).

I want to highlight that the problem is not using SharedValue but repeatedly needing to read it from within the JS thread. The [example diff you shared](https://github.com/PedroBern/react-native-collapsible-tab-view/commit/9378af6a0f95ce59a3eab9ca786baa05bb8039ba#diff-739a99e40f851900557fbc98026c870d8ac3c2da65ec6fc29d60c161effaf20dR240-R248) is a good example of the correct pattern which is to take something from the JS thread and run it in worklet to save CPU cost in the JS thread.

andreialecu:
Are you using the new architecture?

With Paper/old architecture, as far as I remember, calling `SharedValue.value` in the past did not guarantee you'd get the actual last value. It was sometimes out of sync with the value from a worklet. Hence why I needed to create the `useConvertAnimatedToValue` hook, to try to ensure the value is kept in sync as much as possible.

AndreiCalazans:
@andreialecu I'm not using the new arch nor did I test in it.

AndreiCalazans:
@andreialecu note that my changes avoids this interchange of SharedValue to regular React state and solely relies on React state where possible.

src/Container.tsx
```diff
@@ -242,7 +236,7 @@ export const Container = React.memo(
       )
 
       useAnimatedReaction(
-        () => headerHeight.value,
+        () => headerHeight,
```
andreialecu, line 239: Does this actually work right? Can reanimated trigger a reaction based on a react state change?

andreialecu:
> @andreialecu note that my changes avoids this interchange of SharedValue to regular React state and solely relies on React state where possible.

I understand. 

It's just that this transition to animated values was actually done on purpose, and it greatly improved performance. I'm trying to understand what changed recently for this to no longer be the case.

If accessing shared values from JS was indeed such performance-intensive, I think that it would've been discovered by others in the past. I find it odd that this is the first time I'm hearing about this, and I couldn't find other mentions of this potential problem anywhere else.

Reanimated has regressions some times, so it would be great to understand more about this from an upstream maintainer. Perhaps we can ping one on Twitter/X

andreialecu:
What I think would help is if you could create a separate, minimal repro of this SharedValue issue, as a quick benchmark. Then we can perhaps run it on various reanimated versions and see if it's indeed a regression.

AndreiCalazans:
@andreialecu I understand the pushback. It will be helpful to hear from @piaskowyk to see if my understanding is correct. 

I think the Example app is enough of a minimal repro. I [pushed here](https://github.com/PedroBern/react-native-collapsible-tab-view/pull/424) a simple setup that allows us to run performance tests using Flashlight. 

I ran [these e2e](https://github.com/PedroBern/react-native-collapsible-tab-view/pull/424/files) tests with 10 iterations for `main` and for this branch. All tests were ran on the same Samsung a23 device (considered low end). These are the results:

Screenshot 2024-06-19 at 13 02 53

At a first glance it seems like nothing has changed and this branch is just slightly better, but take a look at when the FPS starts dipping when we navigate to the FlashList example. On `main` the dip happens at around 3 seconds for the 10 iterations while on the refactored branch it happens around 2.5 seconds this means this branch is half a second faster to load the components and dispatch the paint to UI natively. 

| Before | After |
|--|--|
| Screenshot 2024-06-19 at 13 10 08 | Screenshot 2024-06-19 at 13 02 53 |


Flashlight reports:
- [Main branch](https://github.com/user-attachments/files/15903609/com.rnctvexample.json)
- [andrei/optimize-tab-view branch](https://github.com/user-attachments/files/15903608/flashlight_report_with_changes.json)

To visualize these yourself install Flashlight and run `flashlight report fileone filetwo`

src/Container.tsx
AndreiCalazans: Yea, it does work, I added logs to validate.

src/hooks.tsx
```diff
@@ -119,6 +118,20 @@ export function useTabNameContext(): TabName {
...
 }
 
+export function useLayoutHeight(initialHeight: number = 0) {
+  const [height, setHeight] = useState(initialHeight)
+
+  const getHeight = useCallback(
+    (event: LayoutChangeEvent) => {
+      const latestHeight = event.nativeEvent.layout.height
+      if (latestHeight !== height) {
+        setHeight(latestHeight)
```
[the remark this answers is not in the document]
nandorojo: Is this check necessary? doesn't react not even re-render if the value is the same?

I believe this is actually an anti-pattern, since now `getHeight` relies on the `height` variable, causing it to update its reference based on its parent state when this isn't necessary.

For example, you could just dispatch it with a callback too, right?

```tsx
setHeight(height => height === latestHeight ? height : latestHeight)
```

However, that becomes the same as just doing `setHeight(latestHeight)`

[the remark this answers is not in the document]
andreialecu: Maybe relevant: https://github.com/facebook/react/issues/28779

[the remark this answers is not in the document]
andreialecu: Also see:
https://legacy.reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update

> Bailing out of a state update
If you update a State Hook to the same value as the current state, React will bail out without rendering the children or firing effects. (React uses the [Object.is comparison algorithm](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#Description).)

> Note that React may still need to render that specific component again before bailing out. That shouldn’t be a concern because React won’t unnecessarily go “deeper” into the tree. If you’re doing expensive calculations while rendering, you can optimize them with useMemo.

The problem here is that all the code in the component will run again unnecessarily, even though it doesn't go "deeper" or actually render anything. By preventing setting the state, we can optimize against that.

andreialecu:
I released this as 8.0.1-rc.0 to npm for the time being, so that we can all test it more easily.

src/Lazy.tsx
```diff
@@ -54,13 +45,24 @@ export const Lazy: React.FC<{
...
   const isSelfMounted = React.useRef(true)
 
-  const opacity = useSharedValue(cancelLazyFadeIn || startMounted.value ? 1 : 0)
+  /**
+   * We start mounted if we are the focused tab, or if props.startMounted is true.
+   */
+  const shouldStartMounted =
+    typeof _startMounted === 'boolean'
+      ? _startMounted
+      : focusedTab.value === name
```
andreialecu, line 60: It seems that reading `.value` in render is not recommended, according to: https://x.com/kzzzf/status/1805222107011317861

This is not your fault though, the code previously did this too.

andreialecu:
Released latest changes as 8.0.1-rc.1

CHANGELOG.md
```diff
@@ -1,314 +1,303 @@
+## [8.0.1](https://github.com/PedroBern/react-native-collapsible-tab-view/compare/v8.0.0...v8.0.1) (2024-07-05)
```
AndreiCalazans, line 1: @andreialecu sorry I ran prettier on this file and it changed formatting. My main changes were just the 8.0.1 entry

PR #461, opened 2025-02-14
fix: new arch flickering

voslartomas:
## Description

This part of code is trying to scroll view in every frame for 1500ms which means hundreds of operations, before new architecture it was working fine, but it seems now with new architecture it causes problems when you switch tabs. I just added simple condition so it will scroll only every 100ms, so much less operations and it seems fine with this fix.


nikitapilgrim:
up

dmaykov:
Patching with this change solved mine issue on New Arch.

varshithmee:
This fixed laggy scrolling for me 👍

rileysay:
@andreialecu can this be merged?

matthieubessol:
This also fixed it on both Android and iOS for me.

andreialecu:
This may be fixed by #483 which is available as v9.0.0-rc.0

PR #483, opened 2025-09-20
Update to RN 0.81, Reanimated 4, FlashList 2

andreialecu:
A few breaking changes here, this requires new architecture and reanimated v4.

Also removes MasonryFlashList because it's now just a prop for FlashList.


andreialecu:
Released to npm as v9.0.0-rc.0 for testing

compojoom:
Thanks for rc.0! I updated to expo 54 and v 8.0.0 was crashign on it in some situations. With v8 it works good!

castrolem:
With v8 and this rc.0 version I'm getting a crash when navigating to the screen that contains the example code with expo router, the screen does a proper render tbh, because I can see the animation happening up to the middle of the screen before the crash. I haven't done a clean project, but I wasn't doing much, just adding a new screen that contained the same code as the example to try it out.

Update: I think this may have been related to Nativewind, I'm investigating it.

Randall71:
I just tested this one and notice that the Flashlist version is a bit slow. 
I think version 2.0.2 is used here, updating to the latest version 2.1.0 would be interesting, what do you think?

ReedYu:
> With v8 and this rc.0 version I'm getting a crash when navigating to the screen that contains the example code with expo router, the screen does a proper render tbh, because I can see the animation happening up to the middle of the screen before the crash. I haven't done a clean project, but I wasn't doing much, just adding a new screen that contained the same code as the example to try it out.
> 
> Update: I think this may have been related to Nativewind, I'm investigating it.
The same question, have you got the results yet?

castrolem:
> The same question, have you got the results yet?

I don't have an aswer @ReedYu, I tried with the rc of nativewind but no luck, and I pivoted because we are changing our UI and I had to write a native module for our use case 😞

Kerumen:
@andreialecu With the [recent release of Expo 55 Beta](https://expo.dev/changelog/sdk-55-beta) and the removal of the Legacy Architecture, do you think we can move this PR forward?

